Issue #254, opened 2013-12-03
Add support for storing the order in which HTTP headers are received.


hellais:
Currently TrueHeaders does not support storing the order in which HTTP headers are received or sent.

This probably requires also changes to the twisted.web.client.Agent.

PR #298, opened 2014-04-04
Modified TrueHeaders class to order headers

jacobhaven:
Fixes [Issue 254](https://github.com/TheTorProject/ooni-probe/issues/254) by storing TrueHeaders in an `OrderedDict`.

Removes clutter from TrueHeaders class by making `_rawHeaders` a simple `OrderedDict` mapping from header string to list of value strings. This conforms with its type in the superclass (`twisted.web.http_headers.Headers`) and removes the need to override a few methods. A new dict `_headerCases` stores the mapping from lowercase header name to their original capitalization. This is accessed by overriding the superclass method `Headers._canonicalNameCaps`.

Adds `test_trueheaders.test_order_preserved` to ensure `getAllRawHeaders` returns headers in the order they were set.

### DIFF
--- a/ooni/tests/test_trueheaders.py
+++ b/ooni/tests/test_trueheaders.py
@@ -36,6 +36,15 @@ class TestTrueHeaders(unittest.TestCase):
         th = TrueHeaders(dummy_headers_dict)
         self.assertEqual(th.getDiff(TrueHeaders(dummy_headers_dict2), ignore=['Header3']), set())
 
-
-
-
+    def test_order_preserved(self):
+        th = TrueHeaders()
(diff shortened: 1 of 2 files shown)


willscott:
This feels like we can close it as obsolete at this point. The referenced class no longer exists in the code base. @hellais - consider closing (+ opening a new bug if the issue remains)

hellais:
@willscott sure, will do. I also think that it's not longer so strictly necessary to preserve the header order for the Agent used by ooni-probe and if this is a desired feature, it should probably be part of measurement-kit, rather than ooni-probe.

PR #338, opened 2014-07-23
Fix some bugs when invoking and using oonid.

NullHypothesis:
- `config.global_options['no-geoip']` failed because of non-existing key.
- The second argument passed to `startNetTest()` was no string.
- `__init__()` in `YAMLReporter` used the current directory instead of the report directory.


hellais:
I am going to close this ticket and consider future work be done on branch: https://github.com/TheTorProject/ooni-probe/pull/442

PR #358, opened 2014-10-22
Feature/new sniffer subsystem

kudrom:
New approach focused on filters for the different templates.


hellais:
This is really good stuff, though I think that given the fact we are moving to measurement-kit as a testing engine it may not be so relevant in the future. I am going for the moment close this PR (I am cleaning up a bit of open pull requests on ooniprobe).

If you want to rebase it and you think it would be important to add this to ooniprobe, feel free to re-open it though 👍

PR #371, opened 2015-01-03
13581: Avoid hacking sys.path in bin/* scripts as that is a security risk.

infinity0:
We want to run these scripts with extra capabilities, using setcap. However, if
the end user can set PYTHONPATH, then they can feed arbitrary code into these
scripts and give extra capabilities to that code, which we should not allow.

Includes a ooniprobe-dev wrapper shell script to make things easier for
developers running directly from the repository. The important thing is that
this script is developer-only and *not* installed on end users' machines.

### DIFF
--- a/bin/oonideckgen
+++ b/bin/oonideckgen
@@ -1,10 +1,5 @@
 #!/usr/bin/env python
-import os
-import sys
 import exceptions
-
-sys.path[:] = map(os.path.abspath, sys.path)
-sys.path.insert(0, os.path.abspath(os.getcwd()))
 
 from twisted.internet import defer, reactor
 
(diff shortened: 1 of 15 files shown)


infinity0:
note: this is necessary for 13602 which would allow all tests to work as non-root

infinity0:
Hey, any news on this?

hellais:
@infinity0 yeah sorry for not attending to this PR sooner. I have been thinking a bit on wether or not this is the best solution and if there is another way to support running ooniprobe from the git repo directly, but I believe this is in fact the best way.

hellais:
So I thought a bit more about this and I think it's an ok change to make.

I rebased your branch and merged it into master.

hellais:
Not sure why github didn't realise that I did in fact merge it. Anyways your commit is in master here: https://github.com/TheTorProject/ooni-probe/commit/3bad9401e657507abea23bb4d650423bad3d3845 (the author of the commit has been kept).

PR #372, opened 2015-01-03
Fix/14086

hellais:


hellais:
I am going to merge this since it contains important fixes.

PR #373, opened 2015-01-14
Add install package libdumpnet-dev

anadahz:
The package libdumpnet-dev is required by pydumpnet


hellais:
I think you have a typo. It's libdum**b**net

anadahz:
oops

PR #375, opened 2015-01-15
Add support for Linux Mint

anadahz:
Add support for Linux Mint, tested on Linux Mint 17, but it should work for
other versions too.


hellais:
Cool. Merging.

PR #376, opened 2015-01-19
Fix/deckgen stall

hellais:


hellais:
I am going to merge this since it includes crucial bug fixes.

PR #378, opened 2015-02-26
Fix typo.

igueths:
I'm new to the project, and while looking for nettest type stuff to work on I stumbled on this typo and thought I'd fix it. As yet I have not found workflow type docs, so if I should be PR'ing against a different branch upstream or something please let me know. Thanks!


hellais:
Awesome thanks for spotting this!

And yes, this is indeed the best workflow for submitting patches. :+1:

PR #379, opened 2015-02-27
Joe/13597

hellais:


hellais:
I tried running this with this set of bridges as input:

```
obfs4 178.209.52.110:443 67E72FF33D7D41BF11C569646A0A7B4B188340DF cert=Z+cv8z19Qb8RxWlkagp7SxiDQN++b7D2Tntowhf+j4D15/kLuj3EoSSGvuREGPc3h60Ofw iat-mode=0
obfs4 83.212.101.3:41213 A09D536DD1752D542E1FBB3C9CE4449D51298239 cert=lPRQ/MXdD1t5SRZ9MquYQNT9m5DV757jtdXdlePmRCudUU9CFUOX1Tm7/meFSyPOsud7Cw iat-mode=0
obfs3 83.212.101.3:80 A09D536DD1752D542E1FBB3C9CE4449D51298239
obfs3 109.105.109.163:38980 1E05F577A0EC0213F971D81BF4D86A9E4E8229ED
```

(these are the bridges include in the Tor browser bundle)

It seems like the obfs4 bridges are failing to be tested on my machine and I don't see any debug log in the ofbs4proxy_log key. Can you see what the issue is?

Also I would suggest we just keep one key for the log (not have obfsproxy_log and obfs4proxy_log).

It may be that it is failing because I have an old version of Tor, but I would still expect to see something in the log output that would help me debug why it is failing and what I can do to make it work properly.

I am investigating what the minimum required Tor version for obfs4proxy is and will update once I know it.

Anyways overall this is very good work, thanks!

hellais:
Note: there also seems to be a problem with the bridge unittest: https://travis-ci.org/TheTorProject/ooni-probe/jobs/52408860#L1605

hellais:
From IRC:

```
13:08 < Yawning> hellais: 0.2.5.x
13:08 < Yawning> if you want to use anyting newer than obfs3
```

We should put checks that validate that Tor is newer than that when running obfs4proxy.

joelanders:
(the previous CI failed with some timeout error in test_http_requests, so I pushed again and it went away)

hellais:
This looks very good. I will soon merge it as soon as I can run it, but the code looks good! :+1:

hellais:
:+1: :+1: merged, thanks!

PR #380, opened 2015-03-01
13602: build a binary of bin/ooniprobe that we can setcap on to avoid running as root

infinity0:
See Makefile inside this commit for description; fixes https://trac.torproject.org/projects/tor/ticket/13602


hellais:
I am going to proceed in merging this as it may be useful to people even if we don't ship it just yet.

PR #381, opened 2015-03-25
Feature/unique identifier + Test helper addresses

hellais:
In here I add 2 new keys to the report data format.

The report_id that is a random 64 character value set by the client.

The test_helpers that contains the addresses of the test helpers used.

I will now issue a PR on ooni-spec to include these changes.


hellais:
https://github.com/TheTorProject/ooni-spec/pull/37

hellais:
I am going to proceed in merging this as nobody complained loudly.

PR #382, opened 2015-03-26
Feature/third party template

hellais:


hellais:
From IRC:

```
17:53 < r2r0> other thing that we did this week was start writing the third party test template to
              support writing the CCT tests: https://github.com/TheTorProject/ooni-probe/pull/382
17:53 -zwiebelbot:#ooni- github-OONI-PR#382: Feature/third party template by hellais
17:54 < r2r0> aagbsn gave me some feedback on it
17:54 < r2r0> regarding the resetting strategy for the timer
17:55 < aagbsn> I suggested not resetting the timer, since it competes with OONI's built in timers
17:55 < aagbsn> we might want to read the configured timers and raise an exception if they conflict
                (if the built-in timers are less than the 3rdparty template)
17:56 < r2r0> ah that was not clear to me previously
17:56 < r2r0> we actually are already supposed to be handling this case
17:56 < r2r0> kudrom implemented it some time ago
17:57 < aagbsn> my diff is just      def outReceived(self, data):
17:57 < aagbsn> -        self.resetTimer()
17:57 < aagbsn> otherwise chatty applications keep resetting the timer until OONI's hard limit comes
                along
17:57 < aagbsn> I proposed launching the 3rdparty apps in setUp instead of the test_ methods
17:58 < aagbsn> so that you can have a test class that launches an application, runs various tests,
                and then everything exits at shutdown
17:58 < r2r0> https://github.com/TheTorProject/ooni-probe/commit/e9dfc89
17:58 < r2r0> here it is
17:59 < r2r0> this needs to be investigated properly
17:59 < r2r0> because it needs to be handled properly
17:59 < r2r0> I will add this part of the chat to the PR
18:00 < aagbsn> well, that case isn't handled
18:00 < aagbsn> because the timer gets reset whenever more output appears
18:00 < r2r0> yes I see what you mean
18:00 < r2r0> it will reset only the local timer, but not the measurement timeout one
18:00 < r2r0> they should both share the same timer
18:00 < aagbsn> ah, that is an interesting idea
18:01 < aagbsn> well, timeout is a failure case usually
18:01 < r2r0> yes, but in some cases it's part of the test
18:01 < r2r0> for example tor, never exits
18:01 < r2r0> so you can't use the exit status to determine if it fails
18:01 < r2r0> you need to use time
18:01 < aagbsn> that's why I think that if you need long lived processes, it shouldn't be launched
                as a measurement
18:02 < r2r0> and you know it has failed, because you get nothing on stdout for a certain amount of
              time
18:02 < aagbsn> ah, I thought that the fail/success would be determined by - scraping messages from
                output that indidcate success from the application, or successfully using the
                application to fetch a resource
18:03 < r2r0> yes, in some cases you can do that
18:03 < r2r0> but in others you need to depend on a timeout
18:03 -!- caber [~woohoo@noc.xs4all.nl] has joined #ooni
18:04  * isabela raises her hand for a quick note (don't want to disrupt the discussion tho)
18:04 < r2r0> as in the tor case, where it will stall if it can't connect to any directory authority
18:04 < r2r0> aagbsn: let's continue discussing this on the PR on github
```

hellais:
Since this is a requirement for the following PR's I am going to merge this.

PR #383, opened 2015-03-27
Handle also unhandled exceptions inside of ooni(deckgen|report|resources...

hellais:
...)

* Exit with the correct exit status


hellais:
Lunar confirms that this works properly. I am merging.

PR #386, opened 2015-04-05
Add meek fronted requests test

anadahz:


ooni/nettests/blocking/meek_fronted_requests.py
```diff
@@ -0,0 +1,74 @@
...
+    ajax.aspnetcdn.com:az668014.vo.msecnd.net
+    a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net
+    """
+    name = "meek fronted requests test"
+    version = "0.0.1"
+
+    usageOptions = UsageOptions
+    inputFile = ['file', 'f', None,
+                  "File containing the DomainName:HostHeader combinations to\
+                  be tested, one per line."]
```
hellais: Where is an example file?

Can we make it so there is a default input for this test so that we don't have to fetch the input?

hellais:
I guess this can be closed, since it's super-seeded by https://github.com/TheTorProject/ooni-probe/pull/387?

PR #389, opened 2015-04-22
low memory handling #13140

0xPoly:
Handles situation where system is running out of RAM ([#13140](https://trac.torproject.org/projects/tor/ticket/13140)). If memory is below the critical value (default 25 MB), no new tasks are allow to start. Memory utilization is re-checked every 20 seconds. This patch uses the psutil library, which is cross-platform but 200-300 kilobytes in size. 

Thanks to r2r0 for the help with this patch.

### DIFF
--- a/data/ooniprobe.conf.sample
+++ b/data/ooniprobe.conf.sample
@@ -51,6 +51,8 @@ advanced:
     report_log_file: null
     inputs_dir: null
     decks_dir: null
+    # Minimum amount of RAM (in MB) required to run tests
+    critical_ram: 25
 tor:
     #socks_port: 8801
     #control_port: 8802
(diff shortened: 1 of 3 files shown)


hellais:
Sorry for such a long delay in reviewing this. I am still a bit torn about what should be done about this since it would mean adding an extra dependency to ooni-probe. Since it's packaged in debian (https://packages.debian.org/wheezy/python-psutil) it shouldn't be a big deal, but I am not yet certain it's a good thing.

I will think further about this and then assess.

hellais:
Hey @0xPoly hope all is well (BTW if you do read it we would love to hear back from you).

Thanks for putting this together.

I think that this si unlikely to land into future versions of ooniprobe as it's adding an extra dependency and since we are moving to using measurement-kit as the testing engine this is not going to be super important in the future.

I am going to close this PR for the moment.

PR #392, opened 2015-04-30
tcp scan tutorial

0xPoly:


docs/source/tutorial.rst
```diff
@@ -5,7 +5,122 @@ Writing a TCP port scanner
...
+Setup
+------
+
+OONI requires some basic information from all tests before running them.
+To provide this information we add the following to our test::
+
+    class TCPPortScan(tcpt.TCPTest):
+        name = 'TCP_port_scan_test'
+        author = 'OONI Contributor <ooni@torproject.org>'
+        version = 0.0.1
```
hellais: This should be a string that is it should be `version = "0.0.1"`

hellais:
I am going to merge this and apply that fix

0xPoly:
good catch! thanks for the fix

PR #393, opened 2015-05-04
Add pyasn1-modules as dependancy in requirements.txt

DonnchaC:


hellais:
I was under the impression this would have been fetched by pyasn1, but I guess it doesn't hurt to add it.

PR #395, opened 2015-05-26
Org queue simplified

hellais:
These are changed proposed by @dantheta to support using ooniprobe for gathering data for @openrightsgroup in:
https://github.com/openrightsgroup/cmp-issues/issues/78


ooni/geoip.py
```diff
@@ -190,34 +190,37 @@ def resolveGeodata(self):
...
+                log.debug("Cannot determine the probe IP address with a traceroute, becase of insufficient priviledges")
+            except:
+                log.msg("Unable to lookup the probe IP via traceroute")
+
+            try:
+                yield self.askGeoIPService()
+                log.msg("Found your IP via a GeoIP service: %s" % self.address)
+                self.resolveGeodata()
+                defer.returnValue(self.address)
+            except Exception, e:
```
hellais: Note to self: clean up this like so it's pep8 compliant.

ooni/nettests/blocking/http_requests.py
```diff
@@ -17,6 +17,9 @@ class UsageOptions(usage.Options):
         ['url', 'u', None, 'Specify a single URL to test.'],
         ['factor', 'f', 0.8,
          'What factor should be used for triggering censorship (0.8 == 80%)']]
+    optFlags = [
+        ["withoutbody","B", "Don't send page body with report"],
```
hellais: Should read: "don't include HTTP response body inside of the report"

ooni/oonicli.py
```diff
@@ -1,10 +1,17 @@
...
+import urlparse
 
 from twisted.python import usage
 from twisted.python.util import spewer
-from twisted.internet import defer
+from twisted.internet import defer, reactor, protocol, task
+
+import pika
+from pika import exceptions
+from pika.adapters import twisted_connection
```
hellais: These imports should be moved down into the `runWithDaemonDirector` function since I think at this stage we don't want to make pika a hard dependency. The import of them should be wrapped in a `try: except ImportError` and print an error message saying that for this means of operation pika is required and should be installed with pip install pika.

ooni/oonicli.py
```diff
@@ -315,3 +329,231 @@ def start():
...
+                sys.exit(1)
+        global_options["annotations"] = annotations
+
+    if global_options['no-collector']:
+        log.msg("Not reporting using a collector")
+        global_options['collector'] = None
+        start_tor = False
+    else:
+        start_tor = True
+
```
hellais: There is a lot of code duplication between this function and `runWithDirectory`. This should be further refactored so that there is not as many.

ooni/reporter.py
```diff
@@ -265,7 +265,7 @@ def writeReportEntry(self, entry):
         request_json = json.dumps(request)
         log.debug("Sending %s" % request_json)
 
-        bodyProducer = StringProducer(json.dumps(request))
+        bodyProducer = StringProducer(request_json)
```
hellais: Nice catch!

setup.py
```diff
@@ -156,7 +156,7 @@ def run(self):
         prefix = os.path.abspath(self.prefix)
         self.set_data_files(prefix)
         self.do_egg_install()
-        self.ooniresources()
+        #self.ooniresources()
```
hellais: Why did you comment out the invocation of ooniresources? Perhaps we should add an option to the setup.py script if the user needs to disable it for some reason. Would that work for you?

ooni/oonicli.py
```diff
@@ -315,3 +329,231 @@ def start():
...
+                ch, method, properties, body = yield queue_object.get()
+                log.msg("Got message")
+                data = json.loads(body)
+                counter += 1
+
+                log.msg("Received %d/%d: %s" %(counter, lifetime, data['url'],))
+                # acknowledge the message
+                ch.basic_ack(delivery_tag=method.delivery_tag)
+
+                d = createDeck(url=data['url'].encode('utf8'))
```
hellais: It appears that the `createDeck` function doesn't return any value, are you sure this is working as expected?

ooni/oonicli.py
```diff
@@ -315,3 +329,231 @@ def start():
...
+        global_options["annotations"] = annotations
+
+    if global_options['no-collector']:
+        log.msg("Not reporting using a collector")
+        global_options['collector'] = None
+        start_tor = False
+    else:
+        start_tor = True
+
+    def createDeck(url=None,filename=None):
```
hellais: This function should probably go inside the `ooni.deck` module and perhaps be a method or be alternative parameters to the constructor of `ooni.deck`.

ooni/oonicli.py
```diff
@@ -315,3 +329,231 @@ def start():
...
+            d.addErrback(director_startup_other_failures)
+            return d
+
+        return start()
+
+
+    finished = defer.Deferred()
+
+    @defer.inlineCallbacks
+    def readmsg(_,channel, queue_object, consumer_tag):
```
hellais: I would perhaps refactor this function into a separate object so that you don't have to use a global counter and lifetime, but they can simply be part of the object attributes.

ooni/oonicli.py
```diff
@@ -315,3 +329,231 @@ def start():
...
+                d = createDeck(url=data['url'].encode('utf8'))
+                # When the test has been completed, go back to waiting for a message.
+                d.addCallback(readmsg, channel, queue_object, consumer_tag)
+            except exceptions.AMQPError,v:
+                finished.errback(v)
+
+
+
+    @defer.inlineCallbacks
+    def runQueue(connection, name, qos):
```
hellais: Same as for the above function.

ooni/reporter.py
```diff
@@ -298,8 +299,13 @@ def createReport(self):
             self.agent = SOCKS5Agent(reactor, proxyEndpoint=proxyEndpoint)
 
         elif self.collectorAddress.startswith('https://'):
-            # XXX add support for securely reporting to HTTPS collectors.
-            log.err("HTTPS based collectors are currently not supported.")
+            # not sure if there's something else it needs.  Seems to work.
+            # Very difficult to get it to work with self-signed certs.
+            self.agent = Agent(reactor)
```
hellais: Be careful with this as some older versions of twisted don't support HTTPS securely. We should evalute this and assess which version of Twisted are acceptable and make it possible to use https securely.

ooni/utils/net.py
```diff
@@ -27,7 +27,6 @@ def buildProtocol(self, addr):
 # agents with largest anonymity set.
 
 userAgents = ("Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.1.7) Gecko/20091221 Firefox/3.5.7",
-              "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3 1 2 like Mac OS X; en-us)"
```
hellais: Why was this removed?

bin/ooniprobequeue
```diff
@@ -0,0 +1,23 @@
...
+from twisted.internet import reactor
+
+# This is a hack to overcome a bug in python
+from ooni.utils.hacks import patched_reduce_ex
+copy_reg._reduce_ex = patched_reduce_ex
+
+code=0
+
+# from ooni.oonicli import run
+# run()
```
hellais: These comments should be removed.

setup.py
dantheta: As far as I could tell, the ooniresources call would download the data files every time the install action was performed, which uses a lot of bandwidth.  We disabled this part of the setup script and downloaded the resources as part of the ansible server setup.  Making the ooniresources call a separate action in setup.py (or using a command flag) would work well.  I think the vagrant setup uses the ooniresources script directly, rather than relying on the install action.

ooni/reporter.py
dantheta: In testing, twisted 13.2 on python 2.7.6 didn't care about a non matching hostname or a self-signed certificate, and twisted 15.0 on python 2.7.3 rejects invalid hostnames and self-signed certificates, so I think we can infer that twisted 15.0 the preferable version.

ooni/utils/net.py
dantheta: During testing we found that the user-agent was randomly selected on a per-request basis.  It was possible for the Tor request and the non-tor request to be sent with different user-agents, and if the target site was using user-agent sniffing to show a mobile-optimized site (or to redirect to a separate mobile site), the http_request test would register a false-positive for censorship.  It seemed easier to remove the iPhone user-agent to remove the ambiguity!

dantheta:
Thanks for taking the time to look over the branch!  I've responded in-place to a couple of the queries.  The other items I'll aim to get fixed up.

Thanks again!

ooni/oonicli.py
dantheta: createDeck finishes on line 488, with return start(), and returns the deferred director.allTestsDone.  We could get rid of the start() nested function entirely, I think, and put the four lines it contains into createDeck itself.

ooni/oonicli.py
dantheta: The global counter is a leftover from a previous approach which required the counter to be available from different scopes.  Sticking it in the global scope was a bit lazy of me.  The request counter was put in to deal with a memory leak (which I never did manage to track down), and may not end up being used in the final version.

ooni/reporter.py
dantheta: Because of the difficulties in getting twisted 15 to work with self-signed certificates, I ended up adding our made-up CA cert to the system's set of trusted certificate authorities.  This is probably a better option than making twisted blindly trust a self-signed certificate.

hellais: I will look more into this as I have not yet checked out twisted 15.

Some time ago I started writing this piece of software: https://github.com/hellais/sslpin to be used perhaps also inside of ooniprobe (though the ambitions could be perhaps even a bit larger).

It implements self authenticating urls where the fingerprint of the cert are inside of the URI itself. You can see an example of how it works here: https://github.com/hellais/sslpin/blob/master/examples/get_url.py.

I don't think it will work with recent versions of twisted, but it's perhaps still some useful code to look at to see how to properly implement cert validation in twisted.

hellais:
Based on the feedback from @dantheta in #406 I am going to close this and consider such PR to superseed this.

ooni/nettests/blocking/http_requests.py
bassosimone: Done in #406.

ooni/oonicli.py
bassosimone: Done in #406.

ooni/oonicli.py
bassosimone: Done in #406.

PR #398, opened 2015-06-30
Feature/add nfqueue traceroute

hellais:
@david415 I think it would be cool to consider integrating this!


david415:
howdy... i've had my golang implementation of "parasitic traceroute" working for months now:
https://github.com/david415/ParasiticTraceroute

however I never did finish the python implementation.

hellais:
This is good stuff. I think there is some potential for getting something similar implement in measurement-kit. As such I opened this ticket here: https://github.com/measurement-kit/measurement-kit/issues/1366

PR #402, opened 2015-07-27
Fix/feature/install

anadahz:
Fix install for Fedora


hellais:
What is this?

anadahz:
Some wrong key presses ;)

PR #403, opened 2015-07-27
Fix Fedora install

anadahz:


hellais:
Could you rebase this on top of feature/install so that it merges cleanly?

hellais:
Isn't this the same as #404?

scripts/install.sh
```diff
@@ -159,27 +168,86 @@ install_obfs4proxy() {
...
       fi
       (
       set -x
-      $sh_c "yum -y install gmp-devel"
+      $sh_c "${yum} -y install gmp-devel"
+=======
+      (
+      set -x
+      $sh_c "yum -y install golang"
+>>>>>>> ed26fa9674ed2898afa35110f515c9fb8e0f6381
```
hellais: Looks like some merge conflict went bad.

anadahz: Indeed my bad I think it's safe to close this PR.

PR #404, opened 2015-07-28
Add support for DNF package manager, add required build dependencies

anadahz:
- Fedora 22 introduces the next upcoming major version of Yum DNF, this commit adds support for dnf package manager which is the default for Fedora 22
 - Add required package dependencies: GeoIP-devel, python-devel libdnet-devel gcc-c++

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,7 +11,6 @@ set -e
 TOR_DEB_REPO="http://deb.torproject.org/torproject.org"
 CLOUDFRONT="no"
 INSTALL_PT="yes"
-PYTHONPATH=$(python -c "import sys; print ':'.join(x for x in sys.path if x)")
 
 # These are the minimum ubuntu and debian version required to use the debian
 # package.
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -160,74 +167,27 @@ install_obfs4proxy() {
...
-}
-
-
-setup_backports() {
-  echo "deb http://ftp.de.debian.org/debian/ ${distro_codename}-backports main" > /etc/apt/sources.list.d/stable.list
-  $sh_c "gpg --keyserver pgpkeys.mit.edu --recv-key A1BD8E9D78F7FE5C3E65D8AF8B48AD6246925553"
-  $sh_c "gpg -a --export A1BD8E9D78F7FE5C3E65D8AF8B48AD6246925553 | apt-key add -"
-  $sh_c "apt-get update"
-}
-
```
hellais: Why did you remove the installation of meek and setup of backports?

scripts/install.sh
```diff
@@ -160,74 +167,27 @@ install_obfs4proxy() {
...
-install_go() {
```
hellais: Why was this removed?

hellais:
The commit message is a lie. You claim to only add support for Fedora, while you are also remove a lot of other things.

anadahz:
Yes, testing OONI PR review eyes, good catch!
I guess it's safe to assume that this PR can be closed.

hellais:
FYI in the future a good option is also to force push to the branch that has an open PR instead of re-opening another one.

PR #407, opened 2015-09-11
Bug: add notes to uninstall/downgrade Twisted via pip

anadahz:
pypi ooniprobe requirements installs the latest version of Twisted Currently
due to a bug in Twisted ooniprobe requires with Twisted versions >=12.2.0 and
<=14.0.0


bassosimone:
This is obviously correct, merging!

PR #409, opened 2015-09-19
Fix/process

juga0:
First two commits fix what seems to be bugs.
Last two commits are needed for psiphon/openvpn tests, but are generic for all the tests.


juga0:
s/Last two commits/Last three commits :)

hellais:
Thanks for this. Don't mind the travis failing. It seems like they did some changes to their infrastructure that render it incompatible with our tests.

I will take care of fixing that in the future and will now review this.

:+1:

hellais:
This looks very good.

Thanks for working on it. Merging :+1:

Issue #104, opened 2013-06-12
Test Specifications


tildelowengrimm:
- Specification of each test helper.
- Specification of each test.
- Human-usable explanations of the semantic output of each test.

PR #410, opened 2015-09-20
Feature/psiphon dev

juga0:
Not ready to merge. Look FIXMEs in psiphon.py, specially lines 68 and 122.
Thanks.

### DIFF
--- a/ooni/errors.py
+++ b/ooni/errors.py
@@ -6,7 +6,7 @@ from twisted.web.error import Error
 from twisted.internet.error import ConnectionRefusedError, TCPTimedOutError
 from twisted.internet.error import DNSLookupError, ConnectError, ConnectionLost
 from twisted.internet.error import TimeoutError as GenericTimeoutError
-from twisted.internet.error import ProcessDone
+from twisted.internet.error import ProcessDone, ConnectionDone
 
 from twisted.python import usage
 
@@ -50,7 +50,8 @@ def handleAllFailures(failure):
         CommandNotSupported,
         ConnectError,
         ConnectionLost,
-        CancelledError)
+        CancelledError, 
+        ConnectionDone)
 
     return failureToString(failure)
 
@@ -147,6 +148,9 @@ def failureToString(failure):
 
     elif isinstance(failure.value, ProcessDone):
         string = 'process_done'
+
+    elif isinstance(failure.value, ConnectionDone):
+        string = 'connection_done'
 
     else:
         # log.err("Unknown failure type: %s" % type(failure.value))
--- a/ooni/nettests/third_party/psiphon.py
+++ b/ooni/nettests/third_party/psiphon.py
@@ -24,10 +24,10 @@ class UsageOptions(usage.Options):
 
 
 class PsiphonTest(httpt.HTTPTest,  process.ProcessTest):
-    
+
     """
     This class tests Psiphon python client
-    
+
     test_psiphon:
       Starts a Psiphon, check if it bootstraps successfully
       (print a line in stdout).
@@ -50,7 +50,7 @@ class PsiphonTest(httpt.HTTPTest,  process.ProcessTest):
         log.debug('PiphonTest._setUp: setting socksproxy')
         self.localOptions['socksproxy'] = '127.0.0.1:1080'
         super(PsiphonTest, self)._setUp()
-        
+
     def setUp(self):
         log.debug('PsiphonTest.setUp')
 
@@ -65,17 +65,16 @@ class PsiphonTest(httpt.HTTPTest,  process.ProcessTest):
         if self.localOptions['psiphonpath']:
             self.psiphonpath = self.localOptions['psiphonpath']
         else:
-            # FIXME: search for pyclient path instead of assuming is in the
-            # home?
-            # psiphon is not installable and to run it manually, it has to be 
+            # psiphon is not installable and to run it manually, it has to be
             # run from the psiphon directory, so it wouldn't make sense to
-            # nstall it in the PATH
+            # install it in the PATH
             from os import path,  getenv
             self.psiphonpath = path.join(
-                getenv('HOME'), 
+                getenv('HOME'),
                  'psiphon-circumvention-system/pyclient')
             log.debug('psiphon path: %s' % self.psiphonpath)
-
+        # psi_client.py can not be run directly because the paths in the
+        # code are relative, so it'll fail to execute from this test
         x = """#!/usr/bin/env python
 from psi_client import connect
 connect(False)
@@ -107,10 +106,14 @@ connect(False)
         self.report['psiphon_installed'] = True
         log.debug("Adding %s to report" % self.report)
 
+        # Using pty to see output lines as soon as they get wrotten in the
+        # buffer, otherwise the test might not see lines until the buffer is
+        # full with some block size and therefore the test would
+        # terminate with error
         finished = self.run(self.command,
-                                    env=dict(PYTHONPATH=self.psiphonpath), 
-                                    path=self.psiphonpath,
-                                    usePTY=1)
+                            env=dict(PYTHONPATH=self.psiphonpath),
+                            path=self.psiphonpath,
+                            usePTY=1)
 
         def callDoRequest(_):
             return self.doRequest(self.url)
@@ -121,7 +124,7 @@ connect(False)
             self.processDirector.transport.signalProcess('INT')
             os.remove(self.command[0])
             return finished
-        
+
         self.bootstrapped.addBoth(cleanup)
         return self.bootstrapped
 
--- a/ooni/nettests/third_party/psiphon.py
+++ b/ooni/nettests/third_party/psiphon.py
@@ -60,7 +60,7 @@ class PsiphonTest(httpt.HTTPTest,  process.ProcessTest):
         else:
             # FIXME: use http://google.com?
             # self.url = 'https://wtfismyip.com/text'
-            self.url = 'https://check.torproject.orggg'
+            self.url = 'https://check.torproject.org'
 
         if self.localOptions['psiphonpath']:
             self.psiphonpath = self.localOptions['psiphonpath']
@@ -112,17 +112,6 @@ connect(False)
                                     path=self.psiphonpath,
                                     usePTY=1)
 
-
-        def addFailureToReport(failure):
-            log.debug("PsiphonTest.test_psiphon.addFailureToReport")
-            log.debug(repr(failure  ))
-            self.report['failure'] = handleAllFailures(failure)
-            self.report['success'] = False
-            log.debug("Adding %s to report" % self.report)
-            # FIXME: these keys are not being wroten in the report
-            # probably because report is being defined in NetTestCase as
-            # a class attribute that is initialized again in NetTescase._setUp
-
         def callDoRequest(_):
             return self.doRequest(self.url)
         self.bootstrapped.addCallback(callDoRequest)
@@ -132,8 +121,7 @@ connect(False)
             self.processDirector.transport.signalProcess('INT')
             os.remove(self.command[0])
             return finished
-            
-        self.bootstrapped.addErrback(addFailureToReport)
+        
         self.bootstrapped.addBoth(cleanup)
         return self.bootstrapped
 
--- a/ooni/nettests/third_party/psiphon.py
+++ b/ooni/nettests/third_party/psiphon.py
@@ -1,0 +1,141 @@
+import tempfile
+import stat
+import os
+
+from twisted.internet import defer, reactor
+from twisted.internet.endpoints import TCP4ClientEndpoint
+from twisted.web.client import readBody
+from twisted.python import usage
+
+from txsocksx.http import SOCKS5Agent
+
+from ooni.errors import handleAllFailures, TaskTimedOut
+from ooni.utils import log
+from ooni.templates import process, httpt
+from ooni.templates.process import ProcessTest
+
+
+class UsageOptions(usage.Options):
+    log.debug("UsageOptions")
+    optParameters = [
+        ['url', 'u', None, 'Specify a single URL to test.'],
+        ['psiphonpath', 'p', None, 'Specify psiphon python client path.'],
+        ['socksproxy', 's', None, 'Specify psiphon socks proxy ip:port.'],]
+
+
+class PsiphonTest(httpt.HTTPTest,  process.ProcessTest):
+    
+    """
+    This class tests Psiphon python client
+    
+    test_psiphon:
+      Starts a Psiphon, check if it bootstraps successfully
+      (print a line in stdout).
+      Then, perform an HTTP request using the proxy
+    """
+
+    name = "Psiphon Test"
+    description = "Bootstraps Psiphon and \
+                does a HTTP GET for the specified URL"
+    author = "juga"
+    version = "0.0.1"
+    timeout = 20
+    usageOptions = UsageOptions
+
+    def _setUp(self):
+        # it is necessary to do this in _setUp instead of setUp
+        # because it needs to happen before HTTPTest's _setUp.
+        # incidentally, setting this option in setUp results in HTTPTest
+        # *saying* it is using this proxy while not actually using it.
+        log.debug('PiphonTest._setUp: setting socksproxy')
+        self.localOptions['socksproxy'] = '127.0.0.1:1080'
+        super(PsiphonTest, self)._setUp()
+        
+    def setUp(self):
+        log.debug('PsiphonTest.setUp')
+
+        self.bootstrapped = defer.Deferred()
+        if self.localOptions['url']:
+            self.url = self.localOptions['url']
+        else:
+            # FIXME: use http://google.com?
+            # self.url = 'https://wtfismyip.com/text'
+            self.url = 'https://check.torproject.orggg'
+
+        if self.localOptions['psiphonpath']:
+            self.psiphonpath = self.localOptions['psiphonpath']
+        else:
+            # FIXME: search for pyclient path instead of assuming is in the
+            # home?
+            # psiphon is not installable and to run it manually, it has to be 
+            # run from the psiphon directory, so it wouldn't make sense to
+            # nstall it in the PATH
+            from os import path,  getenv
+            self.psiphonpath = path.join(
+                getenv('HOME'), 
+                 'psiphon-circumvention-system/pyclient')
+            log.debug('psiphon path: %s' % self.psiphonpath)
+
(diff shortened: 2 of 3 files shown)


juga0:
One FIXME less, line numbers in previous comment doesn't match anymore.

ooni/errors.py
```diff
@@ -6,7 +6,7 @@
 from twisted.internet.error import ConnectionRefusedError, TCPTimedOutError
 from twisted.internet.error import DNSLookupError, ConnectError, ConnectionLost
 from twisted.internet.error import TimeoutError as GenericTimeoutError
-from twisted.internet.error import ProcessDone
+from twisted.internet.error import ProcessDone, ConnectionDone
```
bassosimone: Why `ConnectionDone` indicates the case where you make an HTTP request to a domain that is not reachable?

ooni/errors.py
```diff
@@ -148,6 +149,9 @@ def failureToString(failure):
     elif isinstance(failure.value, ProcessDone):
         string = 'process_done'
 
+    elif isinstance(failure.value, ConnectionDone):
+        string = 'connection_done'
```
bassosimone: If you modify this later you probably need to sync [the specification](https://github.com/TheTorProject/ooni-spec/blob/master/data-formats/df-000-base.md#error-strings).

ooni/nettests/third_party/psiphon.py
```diff
@@ -0,0 +1,129 @@
...
+            self.psiphonpath = path.join(
+                getenv('HOME'), 
+                 'psiphon-circumvention-system/pyclient')
+            log.debug('psiphon path: %s' % self.psiphonpath)
+
+        x = """#!/usr/bin/env python
+from psi_client import connect
+connect(False)
+"""
+        f = tempfile.NamedTemporaryFile(delete=False)
```
bassosimone: Are there any security implications of writing a temporary file and later executing it? Is it possible to obtain the same effect by, say, running psi_client.py?

ooni/nettests/third_party/psiphon.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if self.localOptions['psiphonpath']:
+            self.psiphonpath = self.localOptions['psiphonpath']
+        else:
+            # FIXME: search for pyclient path instead of assuming is in the
+            # home?
+            # psiphon is not installable and to run it manually, it has to be 
+            # run from the psiphon directory, so it wouldn't make sense to
+            # nstall it in the PATH
+            from os import path,  getenv
```
bassosimone: So, in the current scenario the user is expected either to have downloaded psiphon beforehand or to download it in order to run this ooniprobe test, right? If so, I'd say it is enough to search for psiphon in the place where one should have installed it, i.e. the home, unless otherwise specified.

ooni/nettests/third_party/psiphon.py
```diff
@@ -0,0 +1,129 @@
...
+        super(PsiphonTest, self)._setUp()
+        
+    def setUp(self):
+        log.debug('PsiphonTest.setUp')
+
+        self.bootstrapped = defer.Deferred()
+        if self.localOptions['url']:
+            self.url = self.localOptions['url']
+        else:
+            # FIXME: use http://google.com?
```
bassosimone: Yes, I think using google.com is probably better

ooni/nettests/third_party/psiphon.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_psiphon(self):
+        log.debug('PsiphonTest.test_psiphon')
+
+        if not os.path.exists(self.psiphonpath):
+            log.debug('psiphon path does not exists, is it installed?')
+            self.report['success'] = False
+            self.report['psiphon_installed'] = False
+            log.debug("Adding %s to report" % self.report)
+            #FIXME: this completes the test but ooni doesn't stop running, why?
```
bassosimone: We need to figure this out before merging this pull request, IMO

ooni/nettests/third_party/psiphon.py
```diff
@@ -0,0 +1,129 @@
...
+            return defer.succeed(None)
+
+        self.report['psiphon_installed'] = True
+        log.debug("Adding %s to report" % self.report)
+
+        finished = self.run(self.command,
+                                    env=dict(PYTHONPATH=self.psiphonpath), 
+                                    path=self.psiphonpath,
+                                    usePTY=1)
```
bassosimone: What is the implication here of using `usePTY=1`?

(Also, beware of the indent)

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+PSIPHON_PATH=$HOME
+PSIPHON_PYCLIENT_PATH=$PSIPHON_PATH/psiphon-circumvention-system/pyclient
+
+sudo apt-get install mercurial
```
bassosimone: This is very debian specific. I suggest, instead, to run the script through `bash -e`, document that mercurial is needed, and then let the script bail out if `hg` is not found.

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
...
+import json
+
+# Delete 'server_list' if exists
+if os.path.exists("server_list"):
+    # os.remove("server_list")
+    # os.rename("server_list", "server_list")
+    pass
+else:
+    # Download 'server_list'
+    url ="https://psiphon3.com/server_list'
```
bassosimone: Beware that here you need to close the string using " rather than '

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
...
+#./configure
+#make
+#cp ssh ../../../pyclient/
+#cd $PSIPHON_PATH
+
+# create psi_client.dat
+cat <<EOF > psi_generate_dat.py
+#!/usr/bin/env python
+
+import wget
```
bassosimone: wget was not installed by default on my system (MacOS 10.10.5 with homebrew) and I needed to `pip install` it.

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
...
+import os
+import json
+
+# Delete 'server_list' if exists
+if os.path.exists("server_list"):
+    # os.remove("server_list")
+    # os.rename("server_list", "server_list")
+    pass
```
bassosimone: So, here the idea is to actually remove the server_list before merging?

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
...
+    # Download 'server_list'
+    url ="https://psiphon3.com/server_list'
+    wget.download(url)
+
+# convert server_list to psi_client.dat
+dat = {}
+dat["propagation_channel_id"] = "FFFFFFFFFFFFFFFF"
+dat["sponsor_id"] = "FFFFFFFFFFFFFFFF"
+dat["servers"] = json.load(open('server_list'))['data'].split()
+json.dump(dat, open('psi_client.dat', 'w'))
```
bassosimone: Isn't this something that could be done by psiphon code itself?

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
bassosimone: On my system this script is not executable. Is it intended?

bassosimone:
I am currently running into trouble because, when I run the `psiphon` test from inside the virtualenv apparently `pexpect` is not found. This is probably my fault, will check again tomorrow.

In any case, for the records, this is the relevant part of the output:

```
[D] STDOUT: Traceback (most recent call last):\r\n  File "/var/folders/bl/r84rmchd6c5848k6gxvvky_m0000gn/T/tmpBLbQCY", line 2, in <module>\r\n    from psi_client import connect\r\n  File "$HOME/psiphon-circumvention-system/pyclient/psi_client.py", line 22, in <module>\r\n    from psi_ssh_connection import SSHConnection, OSSHConnection\r\n  File "$HOME/psiphon-circumvention-system/pyclient/psi_ssh_connection.py", line 21, in <module>\r\n    import pexpect\r\nImportError: No module named pexpect\r\n
```

juga0:
Briefly: pip install pexpect
I'll comment more on why this is not in the script and all the other comments later.
Thanks a lot for all the detailed comments

bassosimone:
@juga0: actually `pexpect` is installed both in the virtualenv and system-wide:

```
$ find venv -type d -name pexpect
venv/lib/python2.7/site-packages/pexpect
$ find /usr/local/ -type d -name pexpect
/usr/local//lib/python2.7/site-packages/IPython/external/pexpect
/usr/local//lib/python2.7/site-packages/pexpect
```

Must be something else.

bassosimone:
@juga0: sorted out the issue: I had more than one Python installed (the system one and the one installed with brew). Unfortunately `/usr/bin/env` picked up the wrong Python for which `pexpect` was not installed. Fixed by [using `sys.executable` to execute the temp file](https://github.com/bassosimone/ooni-probe/commit/236314cf5aaf87ac835757685e597b370cdf154e), thus using same Python used to run ooniprobe.

bassosimone:
@juga0: Further [updated my diff to use sys.executable](https://github.com/bassosimone/ooni-probe/commit/34c3bac277decb997f65f37aca3c5d2676d78eb8) to avoid removing my Python executable :).

ooni/nettests/third_party/psiphon.py
bassosimone: Actually here ooniprobe terminates after the psiphon test is complete. I've run it several times and it always completed both using existent and nonexistent domains.

ooni/errors.py
bassosimone: Responding to myself, this probably indicates that the proxy closed the connection.

ooni/errors.py
bassosimone: Do we want to keep the failure here as generic as basically "EOF seen" or do we want to be more specific and indicate that the proxy closed the connection?

scripts/psiphon_install.sh
```diff
@@ -0,0 +1,51 @@
...
+dat["servers"] = json.load(open('server_list'))['data'].split()
+json.dump(dat, open('psi_client.dat', 'w'))
+EOF
+
+chmod +x psi_generate_dat.py 
+./psi_generate_dat.py
+cp psi_client.dat $PSIPHON_PYCLIENT_PATH
+
+# assuming to be inside a virtualenv
+pip install jsonpickle pexpect
```
bassosimone: Is there a way to check whether you are inside a virtualenv?

ooni/errors.py
juga0: Exactly

ooni/errors.py
juga0: Ok, i'll add it to the specification.

Maybe it's not a good idea to give a more detailed error message as we don't have enough information on why the error happen.

ooni/nettests/third_party/psiphon.py
juga0: Or check.torproject.org?.

ooni/nettests/third_party/psiphon.py
juga0: Right, ok.

ooni/nettests/third_party/psiphon.py
juga0: It should be ok to write it and execute it as the same user. 
Unfortunately, psi_client.py can not be run directly because most of the paths in the code are relative, so it'll fail to execute from the test.

ooni/nettests/third_party/psiphon.py
juga0: To see output lines as soon as they get wrotten in the buffer, otherwise the test might not see lines until the buffer is full with some block size, and therefore the test would terminate with error.
I'll fix the identation.

scripts/psiphon_install.sh
juga0: No, it's not intented, my fault.
Note for this comment and the rest, that i just added this script so it's easy for others to check the test, but i didn't want it to be merged and i should not add it to this PR, as i'd like to talk first with @anadahz on how the installation script should be done.

scripts/psiphon_install.sh
juga0: Sounds good, i'll check it with @anadahz.

scripts/psiphon_install.sh
juga0: I didn't add wget as it could be install from Debian package *and* pip, other thing to check with @anadahz on which would be the proper way.

scripts/psiphon_install.sh
juga0: Yeah, i think it's better to remove it or overwrite it, as it changes over the time. I commented it out just to accelerate installation while developing.

scripts/psiphon_install.sh
juga0: Ops, will fix.

scripts/psiphon_install.sh
juga0: hasattr(sys, 'real_prefix'). Inside a virtualenv real_prefix points to the prefix of the system Python, outside a virtualenv real_prefix should not exist.

scripts/psiphon_install.sh
juga0: He, probably, but we shouldn't modify the way Psiphon behaves, just test it.
Or do you mean the psiphon test?. If you mean the psiphon test, no, i don't think it should.

ooni/errors.py
bassosimone: I'm proposing something like proxy_connection_eof (or equivalent wording). That is slightly more specific than just plain connection_done. And also more accurate since what happened is an unexpected EOF in the connection to the proxy. Makes sense?

ooni/nettests/third_party/psiphon.py
bassosimone: I've no strong opinion on this. I slightly prefer google.com because accessing google.com conveys less information on one's traffic than accessing torproject.org.

ooni/nettests/third_party/psiphon.py
bassosimone: @juga0: do you still have the problem that ooniprobe hangs?

ooni/nettests/third_party/psiphon.py
bassosimone: cool, thanks!

scripts/psiphon_install.sh
bassosimone: Agreed. Let's concentrate just on the Python part, then!

ooni/errors.py
juga0: Hmm, i guess it also should be add to the specification a test-specs/ts-0xx-psiphon.md file, right?

bassosimone: Yep!

juga0: BTW, @srvetus refactored this file in https://github.com/TheTorProject/ooni-probe/pull/413. If that PR is merged i should update this

scripts/psiphon_install.sh
bassosimone: It was strange that you download something from psiphon own servers and then still need to massage it to provide it to psiphon. So was asking whether is there a psiphon feature or piece of code that could be used to do this operation, rather than writing our own code in the install script.

ooni/errors.py
bassosimone: Ah, thanks for pointing this out! I will check whether I can merge #413 as well.

scripts/psiphon_install.sh
bassosimone: good to know!

ooni/errors.py
juga0: Re. proxy_connection_eof, Twisted returns ConnectionDone, but we don't know whether it's unexpected EOF or not.
I guess we could get ConnectionDone under other circumstances, don't you think?

ooni/nettests/third_party/psiphon.py
juga0: This is the main thing that has to be solved in this test (line #104).
Yes, the test terminates correctly when Psiphon path exists (both using existent and nonexistent domains), but not when the Psiphon path doesn't exist.

ooni/nettests/third_party/psiphon.py
juga0: np :)

scripts/psiphon_install.sh
juga0: ah, i see what you mean. I think there's no other code from Psiphon for linux to do that, as apparently it's done in a different way with the Windows client (see https://groups.google.com/forum/#!topic/psiphon3-developers/cb8CW7Y98nI).

anadahz:
@juga0 is this ready to be tested?

hellais:
Can we merge this? @juga0 @bassosimone

hellais:
@DonnchaC

bassosimone:
@hellais not yet. There is a problem that the test hangs when this test is run using `-ng` options but have not had time to understand this issue yet.

hellais:
@juga0 are you or your assistant on it?

juga0:
Improvements to psiphon install script in https://github.com/juga0/ooni-probe/commits/feature/psiphon-dev-install-script.
Probably that branch should be merged here

juga0:
Psiphon test specifications here https://github.com/TheTorProject/ooni-spec/pull/44

bassosimone:
Merged #425 that was based on this pull request. I have resolved all the issues highlighted in this pull request in a quick-and-dirty way because it was imperative to merge a psiphon MVP into master.

ooni/errors.py
bassosimone: Yes, you're right.

ooni/nettests/third_party/psiphon.py
bassosimone: Got it, thanks!

ooni/nettests/third_party/psiphon.py
bassosimone: For #425 I retained `check.torproject.org`. I think that's good to go. We can always further improve this in the future, if needed.

ooni/nettests/third_party/psiphon.py
bassosimone: In #425 I added 1172083 as a fix for this. It was a quick and dirty fix and perhaps it would be optimal to figure out whether it's OK that a deferred with its callback already called hangs OONI.

scripts/psiphon_install.sh
bassosimone: @juga0 I removed the install script in #425. I think it's best to only include the install script in the lepidopter distribution so we don't have to worry about it being portable here.

PR #411, opened 2015-09-24
OpenVPN Circumvention Test

srvetus:
This test attempts to connect to an OpenVPN server which is specified in an
OpenVPN config file. A HTTP request is sent if the VPN connection is
established successfully.

This basic test should be expanded to parse more of the output from the
OpenVPN client. This output would provide more detailed explaination why
a connection is failing.

### DIFF
--- a/ooni/nettests/third_party/openvpn.py
+++ b/ooni/nettests/third_party/openvpn.py
@@ -3,7 +3,7 @@ from twisted.python import usage
 from twisted.web.client import Agent, readBody
 from ooni.templates.process import ProcessTest
 from ooni.utils import log
-from ooni.errors import handleAllFailures
+from ooni.errors import handleAllFailures, failureToString
 
 import distutils.spawn
 import re
@@ -32,7 +32,7 @@ class OpenVPNTest(ProcessTest):
     name = "OpenVPN Client Test"
     description = "Connects to an OpenVPN server and does a HTTP GET for the specified URL"
     author = "srvetus "
-    version = "0.0.1"
+    version = "0.0.2"
     timeout = 20
     usageOptions = UsageOptions
     requiredOptions = ['url', 'openvpn-config']
@@ -56,6 +56,16 @@ class OpenVPNTest(ProcessTest):
             # OpenVPN needs to be sent SIGTERM to end cleanly
             self.processDirector.transport.signalProcess('TERM')
             self.exited = True
+
+    def processExited(self, reason):
+            """Monkeypatch processExited to log failure if the process ends
+            unexpectedly before OpenVPN bootstraps.
+            """
+            log.debug("Exited %s" % handleAllFailures(reason))
+
+            # Process exited before OpenVPN bootstrapped. Add failure to report
+            if not self.bootstrapped.called:
+                self.bootstrapped.errback(Exception("openvpn_exited_unexpectedly"))
 
     def handleRead(self, stdout=None, stderr=None):
         """handleRead is called with each chunk of data from stdout and stderr
@@ -88,8 +98,8 @@ class OpenVPNTest(ProcessTest):
             self.report['success'] = True
 
         def addFailureToReport(failure):
-            log.debug("failed: %s" % handleAllFailures(failure))
-            self.report['failure'] = handleAllFailures(failure)
+            log.debug("Failed: %s" % failureToString(failure))
+            self.report['failure'] = failureToString(failure)
             self.report['success'] = False
 
         def doRequest(noreason):
@@ -105,6 +115,9 @@ class OpenVPNTest(ProcessTest):
 
         log.debug("Spawning OpenVPN")
         self.d = self.run(self.command)
+
+        # Monkeypatch processExited to log when OpenVPN exits early
+        self.processDirector.processExited = self.processExited
 
         # Try to make a request when the OpenVPN connection successfully bootstraps
         self.bootstrapped.addCallback(doRequest)
--- a/ooni/nettests/third_party/openvpn.py
+++ b/ooni/nettests/third_party/openvpn.py
@@ -57,6 +57,15 @@ class OpenVPNTest(ProcessTest):
             self.processDirector.transport.signalProcess('TERM')
             self.exited = True
 
+    def inConnectionLost(self):
+            """Monkeypatch inConnectionLost to log failure if the process ends
+            unexpectedly before OpenVPN bootstraps.
+            """
+            log.debug("inConnectionLost")
+
+            if not self.bootstrapped.called:
+                self.bootstrapped.errback(Exception("openvpn_exited_unexpectedly"))
+
     def processExited(self, reason):
             """Monkeypatch processExited to log failure if the process ends
             unexpectedly before OpenVPN bootstraps.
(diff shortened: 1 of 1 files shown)


hellais:
This looks like very good code to me. I will test it more properly in the upcoming days.

juga0:
Specification here https://github.com/TheTorProject/ooni-spec/pull/45

hellais:
We need to add support to https://github.com/TheTorProject/lepidopter for the installation script and test this extensively.

@srvetus @juga0 @anadahz can you check this out?

anadahz:
I have tested the OpenVPN test by using a random free VPS service.

I had some issues finding out what wrong with the OpenVPN service.

1. OpenVPN connection failing for whatever reason
2. OpenVPN connection had wrong credentials or none set
In both cases I had a report similar to this:

```
---
input_hashes: []
options: [-c, OpenVPN.config, -u, www.google.com]
probe_asn: AS0
probe_cc: IS
probe_city: null
probe_ip: 127.0.0.1
report_id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
software_name: ooniprobe
software_version: 1.3.1
start_time: 1447876044.0
test_helpers: {}
test_name: openvpn_client_test
test_version: 0.0.1
...
---
{input: null, test_runtime: 56.01319098472595, test_start_time: 1447876052.0}

```

Ideally the full or parts of the OpenVPN fail log should be placed in the
report. Otherwise it's going to be very difficult --almost impossible to
determine the cause of the network interference happened.

Ideally a sample OpenVPN configuration file should be included with required
options for this test to run successfully.

srvetus:
I've revolved the issue with some OpenVPN errors not being logged in 4606063. I was hitting issues `handleAllFailures` not actually handling all failures, but silently raising an exception from `failure.trap()`. This Exception was then causing remaining report processing to be skipped. Ooni really needs to log when  `handleAllFailures` raises an exception or else actually handle ALL failures.

anadahz:
Thanks for revising the OpenVPN. Upon further testing it seems that some
failures are not being handled when are no OpenVPN credentials provided:

```
# ooniprobe -ng ooni-probe/ooni/nettests/third_party/openvpn.py -c ovpn-no-credentials.config -u https://www.google.com
Log opened.
Not reporting using a collector
[D] No test deck detected
[D] Checking if url is present
[D] Checking if openvpn-config is present
You should add annotations for the country, city and ASN
[D] Creating /root/report-openvpn_client_test-2015-11-24T104051Z.yamloo
[D] Writing report with YAML reporter
[D] Starting this task <generator object generateMeasurements at 0x7fdab78c0d20>
[D] Running <class 'openvpn.OpenVPNTest'> test_openvpn_circumvent
[D] Spawning OpenVPN
[D] Checking all tasks for completion 0 == 1
[D] STDOUT: Tue Nov 24 10:40:51 2015 OpenVPN 2.3.4 x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [EPOLL] [PKCS11] [MH] [IPv6] built on Dec  1 2014\n
[D] STDOUT: Tue Nov 24 10:40:51 2015 library versions: OpenSSL 1.0.1k 8 Jan 2015, LZO 2.08\n

Broadcast message from root@ooni-local (Tue 2015-11-24 10:40:51 UTC):

Password entry required for 'Enter Auth Username:' (PID 27941).
Please enter password with the systemd-tty-ask-password-agent tool!

[D] inConnectionLost
[D] errConnectionLost
[D] outConnectionLost
[D] Task <ooni.tasks.Measurement object at 0x7fdab78cbc90> has failed 2 times
Failure: twisted.internet.defer.CancelledError:•
[D] Rescheduling...
[D] Spawning OpenVPN
[D] STDOUT: Tue Nov 24 10:41:19 2015 OpenVPN 2.3.4 x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [EPOLL] [PKCS11] [MH] [IPv6] built on Dec  1 2014\n
[D] STDOUT: Tue Nov 24 10:41:19 2015 library versions: OpenSSL 1.0.1k 8 Jan 2015, LZO 2.08\n

Broadcast message from root@ooni-local (Tue 2015-11-24 10:41:19 UTC):

Password entry required for 'Enter Auth Username:' (PID 27943).
Please enter password with the systemd-tty-ask-password-agent tool!


Broadcast message from root@ooni-local (Tue 2015-11-24 10:41:19 UTC):

Password entry required for 'Enter Auth Username:' (PID 27941).
Please enter password with the systemd-tty-ask-password-agent tool!
```
```
###########################################
# OONI Probe Report for openvpn_client_test (0.0.2)
# Tue Nov 24 10:33:24 2015
###########################################
---
input_hashes: []
options: [-c, ovpn-no-credentials.config, -u, www.google.com]
probe_asn: AS0                                                                                                                                                                                              
probe_cc: ZZ
probe_city: null
probe_ip: 127.0.0.1
report_id: BX3ZQXmpkINup1vnTv9F84wfgz9xQtungogiYQReP9ImGJtJ991LJCm0oFHcvIz1
software_name: ooniprobe
software_version: 1.3.2
start_time: 1448361204.0
test_helpers: {}
test_name: openvpn_client_test
test_version: 0.0.2
...
---
{input: null, test_runtime: 56.01452898979187, test_start_time: 1448361204.0}
...
```

srvetus:
When OpenVPN hangs waiting for user input, the test will be killed by the timer. Ooni will the respawn the test by calling `test_openvpn_circumvent()` on the same instance of the class. 

This leads to unpredictable behaviour as the test state is kept on the class and not in `test_openvpn_circumvent()`. Respawning leads to more callbacks and errbacks being added to the existing callback chains for example.

Is it possible to disable test respawning or force creation of a new instance of the test class?

hellais:
@srvetus what you could do is set the timeout attribute of the test_class so that it will override the default timeout value of ooniprobe as is documented in here: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/tasks.py#L127

hellais:
It should detect when in STDOUT there is a password prompt and if so kill the process and write an error to the ooniprobe report.

srvetus:
@hellais I've tried setting a timeout attribute on the test, how to I override how the timeout is handled so that an error is logged, and the test is not rescheduled. I do not really know how the timeout callback chain works. It would probably be a much more straightforward task for you to complete.

It is not so straightforward as detecting the password prompt on STDOUT. Typically the prompt is output on /dev/tty and not stdout or stderr. On systems running systemd the prompt can also be forwarded to `systemd-tty-ask-password-agent`.

I'm getting a little frustrated with this test as it is taking many more than the billed number of hours. I'm happy to discuss this some more but I think it would be much faster if you or someone who is more familiar with the Ooni architecture could implement the timeout check.

hellais:
I looked a bit into this and it seems like it's non trivial to not re-schedule the measurements without changing the core of ooni-probe. I will add support for this and integrate these changes in your branch.

hellais:
So I merged the changes that @juga0 did and added support for disabling re-scheduling of measurements in ooni-probe. I think now it achieves what we are looking for. @srvetus can you check out this branch: https://github.com/TheTorProject/ooni-probe/tree/openvpn-test-rebased and see if the changes I made make sense to you?

I also disabled the hooking of the closing of the stdin as I believe that behaviour should not necessarily be considered an indication of a failure of the test, instead we will rely only on the timeout to assert if it succeeds or not.

srvetus:
@hellais  thank you for your review and the change to the scheduling interface. The commits are a little difficult to review, but your changes look to make sense.

I've tested your branch against a number of test cases. It appears to be working correctly now and generating report entries successfully. 

LGTM!

PR #412, opened 2015-09-24
Fix developer script

aagbsn:
Fixes the developer script, which does not currently pass arguments or use the correct path.


bassosimone:
Without this diff on Ubuntu:

```
$ ./bin/ooniprobe-dev -s
./bin/ooniprobe-dev: 6: exec: ./bin/ooniprobe: not found
```

With this diff:

```
$ ./bin/ooniprobe-dev -s
Log opened.
# Installed nettests
* TCP Connect (blocking/tcp_connect)
  Performs a TCP connect scan of all the host port combinations given as input.
[snip]
```

Merging!

PR #413, opened 2015-09-24
Refactor error handling to avoid inconsistencies

srvetus:
The error handling code in `handleAllFailures()` and `failureToString()`
is inconsistent and does not correctly trap all of the Failures for
which there are error messages defined.

The revised code places the Exception and error message in a single
place to avoid inconsistencies arising in future.


bassosimone:
I agree with the approach of unifying failure handling. This is cleaner than before!

(And too bad that dictionaries are not ordered :-).

I don't fully understand the implication of also handling `Exception` with `failure.trap()`. How the behavior changes with respect to before? (My understanding is that the intent of the original code was for unknown exceptions to propagate to the next errback in the chain, while here you basically stop error processing. Is this correct? Are there pieces of ooni-probe that make the assumption that `handleAllFailures()` should propagate unknown errors?)

I also have seen that, compared to the previous tuple, in addtion to adding `Exception, you also have:

- moved `ConnectionLost` and `CancelledError` at the beginning

- swapped `NetworkUnreachable` and `ConnectionError`

Since you mention in your commit message that subclass must be considered, I wonder whether issues issues were considered to make these changes.

Speaking of which, I am confused that `socks_error` is in the middle of the list while other apparently more specific socks related errors appear below. Is this because they are not part of the same inheritance chain despite them being related all to socks?

Also I wonder whether there is a way to automatically check that it does not happen that parent classes appear above child classes in the list above.

hellais:
@srvetus can you reply to the questions of @bassosimone. I agree, but it would be to have an answer logged.

hellais:
This now needs to be rebased against the current master.

srvetus:
I've rebased this branch against master. I've also added a unittest to ensure that child/more-specific failures are always listed before less specific parent Failure classes.

I've also removed the catch-all Exception() failure from the list so as not to change the behavior of `handleAllFailures`. It would help developers if uncaught exceptions from `trap()` also produced a log message. It can be unclear that `handleAllFailures` reraises the Failure.

Should I add a log message?

hellais:
This is great stuff. I am going to proceed in merging it.

Thanks a bunch!

PR #416, opened 2015-10-09
Improve ./bin/ooniprobe-dev

bassosimone:
- make sure it also works on MacOSX where `realpath` is not available

- make sure it bails out with an explicit error message if finding out the toplevel directory path is not possible

- make sure arguments are correctly forwarded to `./bin/ooniprobe` [edit: this was already in #412, I just added quotes around $@ since this appears to be even more correct]

Fixes `./bin/ooniprobe-dev -s` on my system.

Spotted while performing the review of #410.


bassosimone:
Uhm, it appears that part of what I did there was also done by @aagbsn in #412!

bassosimone:
Just rewritten my diff to apply on top of changes in #412.

hellais:
Looks good to me!

PR #419, opened 2015-10-15
Add Bridge Reachability Test documentation

anadahz:

PR #420, opened 2015-10-15
Add meek fronted requests documentation

anadahz:


hellais:
Is this not a superset of #419?

anadahz:
No it's documentation for a different test.

hellais:
Though it includes also the first two commits of the other branch.

In future always branch from master.

PR #401, opened 2015-07-27
Fix/twisted15.0

hellais:

PR #424, opened 2015-11-13
Fix regression I introduced in #401

bassosimone:
It turns out that I didn't realized that merging #401 I introduced a regression. With Twisted > 15.0, [txsocksx creates an Agent wrapper](https://github.com/habnabit/txsocksx/blob/110f71d11704e6f9811701fd28424f400f8510b2/txsocksx/http.py#L32). We must detect this case and properly set the `_pool` attribute of the real Agent, not of the wrapper. Otherwise, headers are not TrueHeaders but just Twisted Headers and hence the `getDiff()` method is not defined, as reported by @anadahz in #401. Thanks to @anadahz for detecting the regression and for prodding me to fix it.

### DIFF
--- a/ooni/utils/trueheaders.py
+++ b/ooni/utils/trueheaders.py
@@ -20,6 +20,9 @@ SOCKS5ClientFactory.noisy = False
 
 from ooni.utils import log
 
+import twisted
+from twisted.python.versions import Version
+
 
 class TrueHeaders(http_headers.Headers):
     def __init__(self, rawHeaders=None):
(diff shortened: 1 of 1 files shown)


bassosimone:
I run `./bin/ooniprobe-dev -ng blocking/http_requests -u http://www.wikipedia.org` in different virtualenvs where I installed Twisted 15.4.0, 14.0.0, 13.0.0, 12.3.0. The output that I obtained in all cases did not indicate any error and was like this (showing Twisted 15.4.0 but others are equal):

```
Error Traversing Database for ipnum = 778419959 - Perhaps database is corrupt?
Successfully bootstrapped Tor
You should add annotations for the country, city and ASN
Performing GET request to http://www.wikipedia.org over Tor
Performing GET request to http://www.wikipedia.org
Headers appear to match
Report ID: [snip]
```

anadahz:
@bassosimone awesome! Thanks for fixing this!

PR #427, opened 2015-11-17
Fix/travis

anadahz:
An attempt to fix travis


bassosimone:
I already tested an earlier version of this diff earlier this afternoon, merging!

Issue #439, opened 2015-12-01
Incompatibility with Twisted==15.5.0


andresazp:
On my installs, the current master seems to be incompatible with Twisted==15.5.0.

It does work with Twisted==15.4.0

I'm not sure if this comes as a consequence merge #437. As I was using 1.3.2 with those fixes before twisted 15.5 was on pip

bassosimone:
Yes, I've also seen this in travis-ci when I merged #437. Around the time I merged #437 I had a chat with @hellais in which he confirmed that the issue we're talking about is a txtorcon issue that must be fixed upstream (they were pulling some bits of Twisted that now have been removed from Twisted).

hellais:
When this PR https://github.com/meejah/txtorcon/pull/147 get's merged in txtorcon and a new release is done, we should be all set.

PR #440, opened 2015-12-02
Avoid installing Twisted 15.5 to avoid bugs with txtorcon

hellais:
This is a workaround to fix https://github.com/TheTorProject/ooni-probe/issues/439


bassosimone:
This seems to break Twisted 14.0.1 with Travis. Will restart the build and see what happens. Meanwhile, I will also test this locally.

bassosimone:
Well, the restarted build completed without errors, so I'm going to merge this.

PR #442, opened 2015-12-02
Implement GUI for ooniprobe

hellais:
I am going to open to PR to work on improving oonid. This superseeds: https://github.com/TheTorProject/ooni-probe/pull/338

### DIFF
--- a/ooni/api/spec.py
+++ b/ooni/api/spec.py
@@ -82,17 +82,21 @@ class Inputs(ORequestHandler):
         input_file = self.request.files.get("file")[0]
         filename = input_file['filename']
 
-        if not filename or not re.match('(\w.*\.\w.*).*', filename):
-            raise InvalidInputFilename
+        expected_format = "(\w.*\.\w.*).*"
+        if not filename or not re.match(expected_format, filename):
+            raise InvalidInputFilename("File name should match regular "
+                                       "expression: %s" % expected_format)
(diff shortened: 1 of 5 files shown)


hellais:
I think the problem lies in the code for the app client. I plan to work on refactoring all of the code within the `data/ui` tree using the tools that modern web developers now have at their fingertips. It should make developing this much easier and faster.

0xPoly:
Alright, so there have been a number of developments regarding this PR. I'm just consolidating them here for visibility:
* During [this meeting](http://meetbot.debian.net/ooni/2016/ooni.2016-03-28-17.00.log.txt) we agreed that development efforts should be focused on writing a web GUI for ooniprobe
* We agreed that the best way forward would be a rewrite of the current code, especially keeping in mind reducing dependencies. @hellais suggested that we use the built-in capabilities of twisted.web to implement the web server
* During an earlier meeting, we also considered that this GUI could be used both on desktops but also on the RPi/ledilopter.
* Finally, some mockups can be found [here](https://darkdepths.net/drafts/ooniprobe-gui-mockup.html)

bassosimone:
@hellais can I close this pull request, then?

hellais:
Yeah, sure.

bassosimone:
LOL that coveralls has commented only now on this PR.

PR #443, opened 2015-12-03
We no longer need to avoid twisted 15.5.0

hellais:
Thanks to @meejah for releasing a fix so quickly


bassosimone:
Two automatic builds failed for supposedly transient reasons on travis-ci. I restarted the build and they completed. Going to merge this hotfix!

PR #449, opened 2015-12-20
No need to downgrade Twisted

anadahz:
All Twister version are supported by ooniprobe!
Fixed (https://trac.torproject.org/projects/tor/ticket/15944)


bassosimone:
This diff looks good to me!

PR #445, opened 2015-12-09
Fix to make the psiphon test run

hellais:
* Remove unneeded imports

PR #451, opened 2015-12-22
Fix/psiphon

juga0:
changes to fix issues commented in https://github.com/TheTorProject/ooni-probe/pull/445

### DIFF
--- a/ooni/nettests/third_party/psiphon.py
+++ b/ooni/nettests/third_party/psiphon.py
@@ -116,7 +116,19 @@ connect(False)
                             usePTY=1)
 
         def callDoRequest(_):
-            return self.doRequest(self.url)
+            log.debug("PsiphonTest.callDoRequest: %r" %(_,))
+            d = self.doRequest(self.url)
+            def addSuccessToReport(res):
+                log.debug("PsiphonTest.callDoRequest.addSuccessToReport")
+                self.report['success'] = True
+                return res
+            d.addCallback(addSuccessToReport)
+            def addFailureToReport(res):
+                log.debug("PsiphonTest.callDoRequest.addFailureToReport. res=%r" % (res,))
+                self.report['success'] = False
+                return res
+            d.addErrback(addFailureToReport)
+            return d
         self.bootstrapped.addCallback(callDoRequest)
 
         def cleanup(_):
--- a/ooni/nettests/third_party/psiphon.py
+++ b/ooni/nettests/third_party/psiphon.py
@@ -88,6 +88,7 @@ connect(False)
         log.debug('PsiphonTest.test_psiphon')
 
         self.report['success'] = None
+		self.report['psiphon_installed'] = None
         if not os.path.exists(self.psiphonpath):
             log.err('psiphon path does not exists, is it installed?')
             self.report['psiphon_installed'] = False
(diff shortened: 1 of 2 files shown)


ooni/nettests/third_party/psiphon.py
```diff
@@ -88,8 +88,10 @@ def test_psiphon(self):
         log.debug('PsiphonTest.test_psiphon')
 
         self.report['success'] = None
+		self.report['psiphon_installed'] = None
```
hellais: I think the indentation of this line is off (looks like some tabs are in here instead of spaces).

ooni/nettests/third_party/psiphon.py
```diff
@@ -88,8 +88,10 @@ def test_psiphon(self):
...
         if not os.path.exists(self.psiphonpath):
             log.err('psiphon path does not exists, is it installed?')
+            self.report['success'] = False
```
hellais: I would keep this as None, since this allows us to distinguish between a non installed psiphon and an installed, but not functioning psiphon.

juga0: removed that line as it is initialized as None

ooni/nettests/third_party/psiphon.py
juga0: yes, sorry, used wrong editor, fixed https://github.com/TheTorProject/ooni-probe/commit/2bb4bd95770905c5111179c853417d530e623bba

ooni/templates/process.py
```diff
@@ -103,12 +103,10 @@ def _setUp(self):
 
     def processEnded(self, result, command):
         log.debug("Finished %s: %s" % (command, result))
-        key = ' '.join(command)
-        self.report[key] = {
-            'stdout': result['stdout'],
-            'stderr': result['stderr'],
-            'exit_reason': result['exit_reason']
-        }
+        self.report['command_name'] = ' '.join(command)
```
hellais: This should be inside of a list since the template supports running multiple commands.

juga0: Do you mean a list of dictionaries?, could you write an example output where there would be several several commands, each of them with a list of the key-values you mentioned in https://github.com/TheTorProject/ooni-probe/pull/445#issuecomment-163318796?
Would it be something like:
- { command_exit_reason: process_done, command_name: /home/foo/.virtualenvs/ooniprobedev/bin/python /tmp/tmpXBR39b }
- { command_exit_reason: connection_refused, command_name: /home/foo/.virtualenvs/ooniprobedev/bin/python /tmp/tmpfoo }

hellais:
> Do you mean a list of dictionaries?, could you write an example output where there would be several several commands, each of them with a list of the key-values you mentioned in #445 (comment)?

Yes correct.

It would look something like:

```
commands: [{command_exit_reason: process_done, command_name: /home/foo/.virtualenvs/ooniprobedev/bin/python /tmp/tmpXBR39b, command_stdout: xxx, command_stderr: bar}, {command_exit_reason: process_done, command_name: /home/foo/.virtualenvs/ooniprobedev/bin/python /tmp/tmpXBR39b, command_stdout: xxx, command_stderr: bar}]
```

hellais:
I reviewed this and it works for me. Good job :+1:

PR #452, opened 2015-12-23
fix checking if commands exist, other minor changes

juga0:
* fix checking if commands exist
* only install virtualenv if it is needed to create a new one
* add/change debug ouput


anadahz:
This is the error that I get while running the scirpt:
```mv: cannot move 'ssh' to '../../../pyclient/pyclient/': Not a directory```

PR #453, opened 2016-01-12
Vagrant readme update

ecneladis:


hellais:
Good catch. Thanks for this!

PR #454, opened 2016-01-18
Added a simple tutorial on how to write test helpers.

elationfoundation:
I had to dig around a bit while I was working on writing my own test-helper to identify what I needed to change to make it function on my ooni-backend instance.  Following a request from @hellais that I write up docs while I play with OONI I wrote up this simple test-helper tutorial. 

The scripts in the docs directory lead me to beleive that all .rst files in the source folder are added into the final docs. If this is incorrect and I have to add the rst file into a index or script of some sort let me know.

### DIFF
--- a/docs/source/writing_test_helpers.rst
+++ b/docs/source/writing_test_helpers.rst
@@ -1,0 +1,137 @@
+Writing Test Helpers
+========
+
+OONI test helpers are...
+
+Writing a Censorship Directionality Test Helper
+--------------------------
+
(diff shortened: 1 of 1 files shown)


hellais:
This looks great. I made some minor fixes to it and included the file in the toc. When we re-build the website it will be included. Thank you very much for your contribution :+1:

anadahz:
@elationfoundation Thanks a lot for your work!
I have added some minor fixes at #459.

PR #456, opened 2016-02-01
Feature/json reports

hellais:
This is not yet ready to be merged, but I am going to create this PR to keep track of progress on it.


.travis.yml
```diff
@@ -1,9 +1,5 @@
 language: python
 env:
-  - TWISTED=Twisted==12.2
-  - TWISTED=Twisted==12.3
-  - TWISTED=Twisted==13.0
-  - TWISTED=Twisted==13.1
```
bassosimone: Why have these versions been removed?

ooni/errors.py
```diff
@@ -26,6 +27,8 @@
     (CancelledError, 'task_timed_out'),
     (gaierror, 'address_family_not_supported_error'),
     (DNSLookupError, 'dns_lookup_error'),
+    (DNSNameError, 'dns_name_error'),
+    (DNSServerError, 'dns_server_failure'),
```
bassosimone: (I suppose there is a reason here why DnsServerError maps to dns_server_failure as opposed to dns_server_error)

ooni/nettest.py
```diff
@@ -219,7 +220,8 @@ def testDetails(self):
             'input_hashes': input_file_hashes,
             'report_id': self.reportID,
             'test_helpers': self.testHelpers,
-            'annotations': self.annotations
+            'annotations': self.annotations,
+            'data_format_version': '0.2.0'
```
bassosimone: Note to self: understand why this was added

ooni/nettest.py
```diff
@@ -329,7 +331,7 @@ def setupTestCases(self, test_cases):
         self.testClasses = set([])
         self.testHelpers = {}
 
-        if config.reports.unique_id is True:
+        if config.reports.unique_id is True and not self.reportID:
```
bassosimone: Note to self: understand why this was added.

ooni/nettests/blocking/dns_consistency.py
```diff
@@ -182,12 +189,14 @@ def lookup_details():
...
                 else:
                     log.msg("Reverse lookups do not match")
                     lookup_details()
                     log.msg("tampering: true")
-                    self.report['tampering'][test_resolver] = True
+                    self.report['errors'][test_resolver] = True
+                    self.report['inconsistent'].append(test_resolver)
 
     def inputProcessor(self, filename=None):
         """
```
bassosimone: Note to self: here it appears that the format of the test report has changed.

ooni/oonicli.py
```diff
@@ -315,10 +320,10 @@ def post_director_start(_):
...
             net_test_loader.annotations = global_options['annotations']
 
-            director.startNetTest(net_test_loader,
-                                    global_options['reportfile'],
-                                    collector)
-        return director.allTestsDone
+            yield director.startNetTest(net_test_loader,
+                                        global_options['reportfile'],
+                                        collector,
+                                        global_options['no-yamloo'])
```
bassosimone: Note to self: check what `@defer.inlineCallback` does since I don't remember.

.travis.yml
hellais: For 2 reasons:

1. It's not possible to use HTTPS safely with twisted <= 13.1

2. In this branch we have started using some functions that are only present starting from twisted 13.2 (namely `readBody`)

I don't think this is a big deal given the fact that debian-backports includes twisted 13.2 and maintaining backward compatibility with such old versions of twisted I expect to be increasingly painful in the long run.

ooni/errors.py
hellais: The reason why I chose to go for this mapping instead of the twisted one is that it seemed to be more "universal" since it's the terminology that is used inside of [RFC1035](https://www.ietf.org/rfc/rfc1035.txt) for RCODE 2 ("Server failure - The name server was unable to process this query due to a problem with the name server").

ooni/nettest.py
hellais: This was added in order to be able to distinguish inside of the data pipeline when a report to be parsed is using the new data format (and hence doesn't require being converted) or is using the old data format.

The ones that end up being published in the end will all have this attribute set to 0.2.0, but not when the are inserted first in the pipeline.

ooni/nettest.py
hellais: This was actually a bug that lead to there being a mismatch between the report_id being generated server-side and the one that was submitted by the client.

Basically we want to support having a report_id also when we write to disk, but we don't want the report_id to be ovewritten by the client if the backend sets it.

Overall if you look through the diffs you will see that the code pertaining to handling the reportID is very fragile and convoluted and would need some better refactoring, but for the moment it works ok.

ooni/nettests/blocking/dns_consistency.py
hellais: Correct.

You can see the diff for this on the ooni-spec PR for the data format 0.2: https://github.com/TheTorProject/ooni-spec/pull/49

ooni/oonicli.py
hellais: It's basically a way to replicate sequential code execution although it's actually asynchronous. What a yield inside of a inlineCallback does is it pauses the execution of the function until the deferred has fired, either with a callback or an errback, and resumes execution from the line below it.

The reason why the behaviour here has changed to run each measurement sequentially is that prior to this change we would schedule all the nettests at the same time.
This lead to the creation of a new report for every nettest, but since there is a limited number of slots for executing the measurements it could have happened that the created report timed out in the backend before it all being written.

By running them sequentially we ensure that we only work on one test at any given time and ensure that we keep the reports alive in the backend.

.travis.yml
bassosimone: Ok, thanks for explaining, I agree!

ooni/nettest.py
bassosimone: Ok, thanks!

ooni/nettest.py
bassosimone: Ack!

ooni/nettests/blocking/http_requests.py
```diff
@@ -18,7 +18,7 @@ class UsageOptions(usage.Options):
         ['factor', 'f', 0.8,
          'What factor should be used for triggering censorship (0.8 == 80%)']]
     optFlags = [
-        ["withoutbody","B", "don't include HTTP response body inside of the report"],
+        ["withoutbody", "B", "don't include HTTP response body inside of the report"],
```
bassosimone: nit here

ooni/nettests/manipulation/http_invalid_request_line.py
```diff
@@ -45,8 +45,11 @@ def setUp(self):
     def check_for_manipulation(self, response, payload):
         log.debug("Checking if %s == %s" % (response, payload))
         if response != payload:
+            log.msg("Detected manipulation!")
+            log.msg(response)
             self.report['tampering'] = True
         else:
+            log.msg("No manipulation detected.")
             self.report['tampering'] = self.report['tampering'] | False
```
bassosimone: unrelated with the diff, but: what's the point of `self.report['tampering'] | False`?

ooni/oonicli.py
bassosimone: So, this is basically a pair of coroutines? Like what you can do in python3 with `yield from`?

ooni/report/tool.py
```diff
@@ -47,9 +47,9 @@ def upload(report_file, collector=None, bouncer=None):
     oonib_reporter = OONIBReporter(report.header, collector)
     log.msg("Creating report for %s with %s" % (report_file, collector))
     report_id = yield oonib_reporter.createReport()
+    report.header['report_id'] = report_id
```
bassosimone: This is again to prefer the report identifier selected by the backend as opposed to using the one written inside the report (if any), correct?

ooni/templates/httpt.py
```diff
@@ -122,11 +122,28 @@ def addToReport(self, request, response=None, response_body=None, failure_string
...
+        def _representHeaders(headers):
+            represented_headers = {}
+            for name, value in headers.getAllRawHeaders():
+                represented_headers[name] = value[0]
+            return represented_headers
+
+        def _representBody(body):
+            # XXX perhaps add support for decoding gzip in the future.
+            try:
+                body.replace('\0', '')
```
bassosimone: This statement has no effect because replace() does not change the string inplace but rather returns a modified string. In any case, I am not sure it's a good idea to alter the body. Why would you like to do that?

ooni/templates/httpt.py
```diff
@@ -122,11 +122,28 @@ def addToReport(self, request, response=None, response_body=None, failure_string
...
+            except UnicodeDecodeError:
+                try:
+                    body = unicode(body, 'utf-8')
+                except UnicodeDecodeError:
+                    body = base64Dict(body)
+
         log.debug("Adding %s to report" % request)
         request_headers = TrueHeaders(request['headers'])
-        request_response = {
+        session = {
```
bassosimone: :-)

ooni/errors.py
bassosimone: Okay. You should probably also add those error strings to [data-formats/df-000-base.md](https://github.com/TheTorProject/ooni-spec/blob/master/data-formats/df-000-base.md) in TheTorProject/ooni-spec#49, right?

ooni/reporter.py
```diff
@@ -257,22 +260,55 @@ def validateCollectorAddress(self):
...
-        if isinstance(entry, Measurement):
-            content += safe_dump(entry.testInstance.report)
-        elif isinstance(entry, Failure):
-            content += entry.value
-        elif isinstance(entry, dict):
-            content += safe_dump(entry)
-        content += '...\n'
 
-        url = self.collectorAddress + '/report'
+        url = self.collectorAddress + '/report/' + self.reportID
```
bassosimone: This is because now you "open" the report using `/report/` and then proceed with posting to `/report/$id`, right?

ooni/reporter.py
```diff
@@ -190,14 +190,16 @@ def _write(self, format_string, *args):
 
     def writeReportEntry(self, entry):
         log.debug("Writing report with YAML reporter")
-        self._write('---\n')
+        content = '---\n'
         if isinstance(entry, Measurement):
-            self._write(safe_dump(entry.testInstance.report))
-        elif isinstance(entry, Failure):
-            self._write(entry.value)
```
bassosimone: You have removed the possibility of serializing `Failure`s. I don't clearly understand the impact of this, but probably is a good idea to raise an exception rather than attempting to serialize a failure.

data/ooniprobe.conf.sample
```diff
@@ -22,6 +22,7 @@ reports:
     # This is a prefix for each packet capture file (.pcap) per test:
     pcap: null
     collector: null
+    format: yaml
```
bassosimone: Where is this used?

bassosimone:
- [ ] finish diff's review
- [ ] test running specific tests
- [ ] test running deck

(does it make sense to test this *before* TheTorProject/ooni-backend#62 is merged?)

ooni/templates/httpt.py
hellais: Oh my you are right that is a very good point.

The rational behind this is that if we end up with bodies with null bytes inside of them this will lead to some JSON parsers from incorrectly parsing the data. Since it's not a big loss to loose null characters this is being added here.

I guess it should probably happen after we have determined that it's not 'ascii' or 'unicode' though since a null byte may actually mean something in a non ascii or utf-8 encoding.

I will push a fix to this.

bassosimone: Shouldn't the parser encoding the original json encode the null byte as `\\u0000`?

To test this hypothesis, here's what I tried:

```
$ python
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 
[GCC 5.2.1 20151010] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> s = "helo\0admin"
>>> s
'helo\x00admin'
>>> json.dumps(s)
'"helo\\u0000admin"'
>>>

$ python3
Python 3.4.3+ (default, Oct 14 2015, 16:03:50) 
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> s = "helo\0admin"
>>> s
'helo\x00admin'
>>> json.dumps(s)
'"helo\\u0000admin"'
>>>

$ nodejs
> s = "helo\0admin"
'helo\u0000admin'
> JSON.stringify(s)
'"helo\\u0000admin"'
>
```

ooni/errors.py
hellais: Yeah added them. Thanks for spotting this.

ooni/nettests/manipulation/http_invalid_request_line.py
hellais: Yeah this is actually sort of stupid. I pushed a fix for it.

ooni/report/tool.py
hellais: Truth

data/ooniprobe.conf.sample
hellais: I had originally added this to support also continued reporting in YAML, but forgot to remove it in the latest version. Good catch.

ooni/oonicli.py
bassosimone: Responding to myself: basically ooni has two schedulers, one for tests and one for reports. A test is bound to lots of measurements when you run a deck. Hence, opening all the reports in parallel could lead to timeouts. This changes how ooni works such that only one test at a time is started (but still measurements in the context of such test go in parallel).

hellais:
@bassosimone I had to add some changes to support the test_start_time using the new format in the backend as well.

bassosimone:
@hellais, I think that a regression was introduced after fdd18a6. If I run a deck without the last two commits included, it works smoothly. Otherwise, I see this:

```
$ ./bin/ooniprobe-dev -c httpo://244z5aoykimasgxi.onion -i deck-it/0.1.0-it-user.deck
Performing GET request to http://anonymouse.org over Tor
Unhandled error in Deferred:
Unhandled Error
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 317, in addCallback
    callbackKeywords=kw)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 306, in addCallbacks
    self._runCallbacks()
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/reporter.py", line 493, in execute
    d = defer.maybeDeferred(f, *arg, **kw)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/reporter.py", line 560, in _incomplete
    raise errors.ReportNotCreated()
ooni.errors.ReportNotCreated:
```

Just by reading the stack trace it's not immediately obvious to me what's wrong. Will read again the diff.

bassosimone:
Okay. Running again with `-v` it's easier to understand what is going wrong (only relevant bits of log attached):

```
[D] Writing report with OONIB reporter
[D] Task <ooni.tasks.ReportEntry object at 0x10d3c0fd0> has failed 2 times
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/tasks.py", line 42, in start
    self._running = defer.maybeDeferred(self.run)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/tasks.py", line 157, in run
    return self.reporter.writeReportEntry(self.entry)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 1274, in unwindGenerator
    return _inlineCallbacks(None, gen, Deferred())
  File "/Users/sbasso/src/TheTorProject/ooni-probe/venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 1128, in _inlineCallbacks
    result = g.send(result)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/reporter.py", line 312, in writeReportEntry
    'content': self.serializeEntry(entry, serialisation_format)
  File "/Users/sbasso/src/TheTorProject/ooni-probe/ooni/reporter.py", line 279, in serializeEntry
    'measurement_start_time': entry.testInstance.report.pop('measurement_start_time', None),
exceptions.AttributeError: 'dict' object has no attribute 'testInstance'
```

bassosimone:
Off we go! This is what I have tested so far. In general I've checked that a looking-good file was written inside the `data/archive` or `data/reports` directory on the server side after a test.

# json

- [ ] tcp-connect
- [ ] traceroute
- [ ] bridge_reachability
- [ ] http_header_field_manipulation
- [ ] meek_fronted_requests
- [ ] dns_consistency
- [ ] http_requests
- [ ] captiveportal
- [ ] dns_spoof
- [ ] daphne
- [ ] http_host
- [ ] http_invalid_request_line

# yaml

- [ ] tcp-connect
- [ ] traceroute
- [ ] bridge_reachability
- [ ] http_header_field_manipulation
- [ ] meek_fronted_requests
- [ ] dns_consistency
- [ ] http_requests
- [ ] captiveportal
- [ ] dns_spoof
- [ ] daphne
- [ ] http_host
- [ ] http_invalid_request_line

hellais:
Thanks a lot for the excellent work on reviewing this.

I added checks for the required options of the tests you pointed out.

Moreover you were right that `SOA` queries were not being serialised properly in the captive portal test and a fix for that has been implemented.

Regarding daphne, I declassed that test to being **experimental** since how it's currently written it's very hard to use and we shouldn't encourage it's usage by average users.

bassosimone:
Hi @hellais!

1) I've done some more testing and noticed one weird thing. If I run, say, http-invalid-request-line using the `yaml` format, my autonomous system number ends up into the results correctly and hence the resulting file name *on the server* also has the correct ASN. This does not happen with the new `json` format where I am always assigned `ASN0` both in the results and hence in the file name on the server. Is this intended and/or expected?

2) Speaking of the file, I reckon that client side the file is named `.yamloo` because it's still yaml. Is it something that it makes sense on server side where we save a json? I'd say yes only if the final file is still valid yaml (a thing of which I am not sure). What do you think?

hellais:
> 1) I've done some more testing and noticed one weird thing. If I run, say, http-invalid-request-line using the yaml format, my autonomous system number ends up into the results correctly and hence the resulting file name on the server also has the correct ASN. This does not happen with the new json format where I am always assigned ASN0 both in the results and hence in the file name on the server. Is this intended and/or expected?

Very good call. I implemented a fix for this in 	29e63ed. It's not exactly the cleanest solution and I think overall there needs to be a lot of refactoring happening in all these classes, but for the moment it works.

> 2) Speaking of the file, I reckon that client side the file is named .yamloo because it's still yaml. Is it something that it makes sense on server side where we save a json? I'd say yes only if the final file is still valid yaml (a thing of which I am not sure). What do you think?

The reason for keeping the client side file in YAML format is that it's easier to skim through manually than if it were JSON. I have verified that the client side YAML file is in fact valid and that when you upload it to the collector with oonireport upload it get's converted properly to JSON as expected.

I would say for the scope of this PR we keep it as YAML and consider dropping YAML support in the client when we have a nice GUI to present the results to the user.

bassosimone:
>> 2) Speaking of the file, I reckon that client side the file is named .yamloo because it's still yaml. Is it something that it makes sense on server side where we save a json? I'd say yes only if the final file is still valid yaml (a thing of which I am not sure). What do you think?
>
> The reason for keeping the client side file in YAML format is that it's easier to skim through manually than if it were JSON. I have verified that the client side YAML file is in fact valid and that when you upload it to the collector with oonireport upload it get's converted properly to JSON as expected.

Yes, I can also confirm this!

> I would say for the scope of this PR we keep it as YAML and consider dropping YAML support in the client when we have a nice GUI to present the results to the user.

I agree. There is value in YAML being so readable.

Perhaps my question was not so clearly written. I was asking why the name of files uploaded on the server (i.e. files in the `data/archive` directory) ends in `.yamloo` when the format of them is now json and not yaml anymore.

(Re the new commit, will review and test it soon!)

hellais:
> Perhaps my question was not so clearly written. I was asking why the name of files uploaded on the server (i.e. files in the data/archive directory) ends in .yamloo when the format of them is now json and not yaml anymore.

This is most probably due to the fact that you have in your `oonib.conf` file the key `report_file_template` set to something with a hardcoded extension. Looking at the example configuration file it seems like it was not updated to reflect the new default format, so I will push a fix for this there.
For reference the default should be `{iso8601_timestamp}-{test_name}-{report_id}-{probe_asn}-{probe_cc}-probe-0.2.0.{ext}`

bassosimone:
# Compare results in json format to specification

- [ ] tcp-connect
- [ ] traceroute
- [ ] bridge_reachability
- [ ] http_header_field_manipulation
- [ ] meek_fronted_requests
- [ ] dns_consistency
- [ ] http_requests
- [ ] captiveportal
- [ ] dns_spoof
- [x] daphne (*skipped*: not relevant anymore)
- [ ] http_host
- [ ] http_invalid_request_line

bassosimone:
Okay, I think this branch was tested and reviewed for long enough and that we should merge this and proceed with further testing with this branch integrated into master.

Specific things to be done later include:

- [ ] make sure that json files match the specification (there is TheTorProject/ooni-spec#49 for this)

- [ ] understand why it was not possible to run scapy based tests

BTW, the branch has conflicts that need to be solved. @hellais could you solve them such that we can merge?

PR #457, opened 2016-02-01
Feature/web connectivity

hellais:
This is the implementation of what is defined inside of https://github.com/TheTorProject/ooni-spec/pull/48. It is not yet ready for merge, but I am creating the PR to keep track of progress on it.


hellais:
I think it would be good to begin testing this branch to spot bugs.

For this purpose I have setup a test bouncer on `httpo://riggrf3c33dnnzuu.onion` that is configured with the web_connectivity test helper.

To run ooniprobe with the web_connectivity test you can do:

```
PYTHONPATH=`pwd` ./bin/ooniprobe -b httpo://riggrf3c33dnnzuu.onion blocking/web_connectivity --url http://torproject.org/
```

anadahz:
**branch**: feature/web_connectivity
**ooniprobe.conf**: `start_tor: false` and `socks_port: 9050`.

`$ ooniprobe -b httpo://riggrf3c33dnnzuu.onion blocking/web_connectivity --url ic-lab.ru`

```
D] Checking if backend is present
[D] Checking if dns-discovery is present
[D] Tor is not running. Skipping IP lookup via Tor.
Looking up your IP address via ubuntu
[D] Got this backend error message {u'error': 404}
Setting collector and test helpers for web_connectivity
Creating report with OONIB Reporter. Please be patient.
This may take up to 1-2 minutes...
```
-----------------------------
**branch**: feature/web_connectivity
**ooniprobe.conf**: `start_tor: false` and `socks_port: 9050`.

`ooniprobe blocking/http_requests --url ic-lab.ru`
```
[D] No test deck detected
[D] Tor is not running. Skipping IP lookup via Tor.
Looking up your IP address via torproject
[D] Got this backend error message {u'error': 404}
Setting collector and test helpers for http_requests
Creating report with OONIB Reporter. Please be patient.
```
--------------------
**branch**: master
**ooniprobe.conf**: `start_tor: false` and `socks_port: 9050`.
`ooniprobe -b httpo://riggrf3c33dnnzuu.onion blocking/http_requests --url ic-lab.ru`

```
[D] No test deck detected
Starting factory <twisted.internet.endpoints.OneShotFactory instance at 0x7fe0cf637d40>
[D] Tor is not running. Skipping IP lookup via Tor.
Looking up your IP address via torproject
Stopping factory <twisted.internet.endpoints.OneShotFactory instance at 0x7fe0cf637d40>
[D] Looking up [{'test-helpers': [], 'version': '0.2.5', 'name': 'http_requests', 'input-hashes': []}]
Setting collector and test helpers for http_requests
Reporting http://ypjon6qqntq6buxz.onion/report
```
--------------------------
The `{u'error': 404}` I guess is  a misconfiguration of the backend? Though it occurs only on the feature/web_connectivity branch.

I don't get though the: `[D] Tor is not running. Skipping IP lookup via Tor.`
```
$ torsocks curl -kis http://wtfismyip.com/text
HTTP/1.1 200 OK
X-Hire-Me: clint@wtfismyip.com
Content-Type: text/plain; charset=utf-8
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
Content-Length: 13
Date: Tue, 17 May 2016 20:56:46 GMT
Connection: keep-alive
```

hellais:
That "error"  `{u'error': 404}` is actually not a real error, but it's a somewhat confusing debug message that appears when you have debug logs enabled.

In this branch I introduced the possibility to test accessibility of collectors and test helpers. To do so I test if we can reach certain paths that don't exist on the backend (`/invalidpath`) and ensure that the error message returned is `{'error': 404}`. Since in the past this was always considered an error and not an expected message it is logged.

I should probably remove logging from inside of that function and log the message only when it's critical.

The other error message `Tor is not running. Skipping IP lookup via Tor.` is also one of these confusing debug log messages. What it should say is actually "Tor control port is not configured, skipping IP lookup via Tor".
IP lookup via Tor (that is via the Tor Control port) can only be done when ooniprobe is instructed to start tor or has a connection to the control port since it uses a control port command to get the IP address of the client via a directory authority.

ooni/deck.py
```diff
@@ -171,11 +166,122 @@ def setup(self):
...
+            yield self.lookupCollectorAndTestHelpers()
+
+
+    def sortAddressesByPriority(self, priority_address, alternate_addresses):
+        onion_addresses= []
+        cloudfront_addresses= []
+        https_addresses = []
+        plaintext_addresses = []
+
+        if priority_address.startswith('httpo://'):
```
willscott: can't this be determined from hostname?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+
+from ooni.backend_client import WebConnectivityClient
+
+from ooni.utils.net import StringProducer, BodyReceiver
+from ooni.templates import httpt, dnst
+from ooni.errors import failureToString
+
+REQUEST_HEADERS = {
+    'User-Agent': ['Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, '
+                   'like Gecko) Chrome/47.0.2526.106 Safari/537.36'],
```
willscott: this is already old. how do we keep it current in a reasonable way?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+                settings=self.localOptions['backend']
+            )
+        else:
+            self.web_connectivity_client = WebConnectivityClient(
+                self.localOptions['backend']
+            )
+
+    def experiment_dns_query(self):
+        log.msg("* doing DNS query for {}".format(self.hostname))
+        return self.performALookup(self.hostname)
```
willscott: Does this / can this include the cname's encountered as well as the final IP's returned?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+            )
+            assert len(answers) > 0
+            assert len(answers[0]) > 0
+            cls.resolverIp = answers[0][0].payload.dottedQuad()
+        except Exception as exc:
+            log.exception(exc)
+            log.err("Failed to lookup the resolver IP address")
+
+
+    def inputProcessor(self, filename):
```
willscott: if we're reading in multiple urls, 2 things that would be nice to have parameterized / be able to do:
1. randomize the order urls are tested
2. add a delay between tests to let the network recover

ooni/deck.py
hellais: Yes, I guess we can add support for both, but we need to support the `httpo` schema since it's used everywhere in ooniprobe to specify tor hidden service addresses.

ooni/nettests/blocking/web_connectivity.py
hellais: Currently this doesn't track CNAME records. I will add support for that as well.

ooni/nettest.py
```diff
@@ -567,12 +566,18 @@ def makeMeasurement(self, test_instance, test_method, test_input=None):
...
+    def initialize(self):
+        for test_class, _ in self.testCases:
+            # Initialize Input Processor
             test_class.inputs = yield defer.maybeDeferred(
                 test_class().getInputProcessor
             )
 
+            # Run the setupClass method
+            yield defer.maybeDeferred(
+                test_class.setUpClass
```
bassosimone: Why a few lines above it's `test_class().getInputProcessor` and here it's `test_class.setUpClass` (i.e. why in the former case `test_class` is invoked and in the latter not)?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+    except AddressValueError:
+        return None
+
+class WebConnectivityTest(httpt.HTTPTest, dnst.DNSTest):
+    """
+    Web connectivity
+    """
+    name = "Web connectivity"
+    description = ("Performs a HTTP GET request over Tor and one over the "
+                  "local network and compares the two results.")
```
bassosimone: It seems to me that web_connectivity does more than that!

bassosimone: Yeah, especially because below the test declares that it does not use Tor

ooni/nettests/blocking/web_connectivity.py
bassosimone: 👍 for both! Perhaps they could be added as part of another pull request?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+                yield row[0]
+
+        def simple_file_generator(fh):
+            for line in fh:
+                l = line.strip()
+                # Skip empty lines
+                if not l:
+                    continue
+                # Skip comment lines
+                elif l.startswith('#'):
```
bassosimone: Strictly speaking `if` would suffice here since above you `continue`

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+            fh = open(filename)
+            line = fh.readline()
+            # Detect the line of the citizenlab input file
+            if line.startswith("url,"):
+                generator = csv_generator(fh)
+            else:
+                fh.seek(0)
+                generator = simple_file_generator(fh)
+            for i in generator:
+                if not i.startswith("http"):
```
bassosimone: Perhaps stupid comment: could it happen that a censored URL is something like `httpfoo.example.com` and then you do not transform it into a URL and then (if I understand it correctly) the test fails for such URL?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+                    i = "http://{}/".format(i)
+                yield i
+        finally:
+            fh.close()
```
bassosimone: Shouldn't this happen in any case when `fh` goes out of scope?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+        else:
+            self.web_connectivity_client = WebConnectivityClient(
+                self.localOptions['backend']
+            )
+
+    def experiment_dns_query(self):
+        log.msg("* doing DNS query for {}".format(self.hostname))
+        return self.performALookup(self.hostname)
+
+    def tcp_connect(self, socket):
```
bassosimone: Shouldn't this be named `experiment_tcp_connect()` for consistency?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+        )
+        self.report['control'] = self.control
+
+    def experiment_http_get_request(self):
+        return self.doRequest(self.input, headers=REQUEST_HEADERS)
+
+    def compare_headers(self, experiment_http_response):
+        count = 0
+        control_headers_lower = {k.lower(): v for k, v in
+                self.report['control']['http_request']['headers'].items()}
```
bassosimone: Is this a compact way to create a `dict`? If so, wow, I didn't know about it!

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,512 @@
...
+             self.report['headers_match'] is True)
+            and self.report['status_code_match'] is True
+        )
+
+        if (dns_consistent == True and tcp_connect == False and
+                experiment_http_failure is not None):
+            blocking = 'tcp_ip'
+
+        # XXX we may want to have different codes for these two types of
+        # blocking
```
bassosimone: Yes, I think that it would be optimal if every branch was labelled differently

ooni/report/cli.py
```diff
@@ -20,7 +20,7 @@ class Options(usage.Options):
          "Specify the configuration file to use."],
         ["collector", "c", None,
          "Specify the collector to upload the result to."],
-        ["bouncer", "b", "httpo://nkvphnp3p6agi5qq.onion",
+        ["bouncer", "b", None,
```
bassosimone: Perhaps stupid question: so, this means that the default is changed, right? Has this implications?

bassosimone:
TODO list for /me:

- [ ] first reading of the diff
- [ ] second reading of the diff
- [ ] testing

ooni/nettests/blocking/web_connectivity.py
hellais: The issue with this is that we need to ensure that it's also kept in sync with the user-agent that is used by the backend.

I think as long as it's not something that is insanely old it should be fine and we can just update them every year or so.

ooni/nettest.py
hellais: Because `getInputProcessor` is an instance method, while `setUpClass` is a class method. The former operates on an instance of `test_class`, while the later operates on the `test_class` directly.

ooni/nettests/blocking/web_connectivity.py
hellais: Oops, this is the description from the `http_requests` test. Need to update it. 👍

ooni/nettests/blocking/web_connectivity.py
hellais: True.

ooni/nettests/blocking/web_connectivity.py
hellais: Fair point. Will make schema detection a bit more strict.

ooni/nettests/blocking/web_connectivity.py
hellais: Technically you are correct. It will be destroyed as part of the `__del__` call as soon as the reference count for it goes to zero. However if an exception is thrown before it goes out of scope, then a reference to it will be kept as part of the stack trace and hence it will not be closed. So it's more robust to generally explicitly call close on it.

ooni/nettests/blocking/web_connectivity.py
hellais: Yeah, magic python one-liners :P

ooni/nettests/blocking/web_connectivity.py
hellais: Sounds reasonable, then maybe one can be `http-blockpage` and the other can be `http-failure`?

ooni/report/cli.py
hellais: Yes this is true, but it was actually a bug in the previous version of oonireport. By having the default set to the canonical bouncer meant that if somebody used oonireport to upload incomplete reports that had been originally submitted to a custom collector (for example because they ran `ooniprobe -c httpo://mycustomcollector`) instead of submitting them to that collector we would be submitting them to our collector.

ooni/nettests/blocking/web_connectivity.py
hellais: Yes I think these are very neat features to have, but I would add these to a separate PR as I think it should be something that is general and not only for this test in particular.

For the URL randomisation we have this ticket: https://github.com/TheTorProject/ooni-probe/issues/480.

I will create one for adding delay, backoff type logic.

ooni/utils/net.py
```diff
@@ -1,23 +1,24 @@
+import re
```
bassosimone: Isn't this import unused?

ooni/templates/httpt.py
```diff
@@ -1,23 +1,28 @@
...
-from ooni.utils.net import BodyReceiver, StringProducer, userAgents
-from ooni.utils.trueheaders import TrueHeaders
+from ooni.utils.net import StringProducer, userAgents
+from ooni.common.txextra import TrueHeaders
+from ooni.common.txextra import FixedRedirectAgent, TrueHeadersAgent
+from ooni.common.http_utils import representBody
 from ooni.errors import handleAllFailures
 
-META_CHARSET_REGEXP = re.compile('<meta(?!\s*(?:name|value)\s*=)[^>]*?charset\s*=[\s"\']*([^\s"\'/>]*)')
+META_CHARSET_REGEXP = re.compile('<meta(?!\s*(?:name|value)\s*=)[^>]*?charset\s*=[\s"\']*([^\s"\'/>!;]+)')
```
bassosimone: A similar regexp is present in `ooni/common/http_utils.py` except that in that file the regexp has not been updated as it has been updated in the change before. It'd better to use just one definition of the regexp.

ooni/backend_client.py
```diff
@@ -0,0 +1,369 @@
...
+                                 query=request)
+
+
+    def closeReport(self, report_id):
+        return self.queryBackend('POST', '/report/' + report_id + '/close')
+
+class WebConnectivityClient(OONIBClient):
+    def isReachable(self):
+        # XXX maybe in the future we can have a dedicated API endpoint to
+        # test the reachability of the collector.
```
bassosimone: I think this command does not apply to the web connectivity client, which has `/status`, right?

ooni/backend_client.py
```diff
@@ -0,0 +1,369 @@
...
+            # We should never be getting an acceptable response for a
+            # request to an invalid path.
+            return False
+
+        @d.addErrback
+        def err(failure):
+            failure.trap(Error)
+            if failure.value.status == '404':
+                return True
+            return False
```
bassosimone: This can be written as `return failure.value.status == '404'`

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,546 @@
...
+    ]
+
+
+def is_public_ipv4_address(address):
+    try:
+        ip_address = IPv4Address(address)
+        if not any([ip_address.is_private,
+                    ip_address.is_loopback]):
+            return True
+        return False
```
bassosimone: Perhaps this could be written as `return not any([...])`

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def compare_titles(self, experiment_http_response):
+        experiment_title = extractTitle(experiment_http_response.body).strip()
+        control_title = self.control['http_request']['title'].strip()
+        first_exp_word = experiment_title.split(' ')[0]
+        first_ctrl_word = control_title.split(' ')[0]
+        if len(first_exp_word) < 5:
+            # We don't consider to match words that are shorter than 5
+            # characters (5 is the average word length for english)
+            return False
```
bassosimone: Perhaps, if the first word is shorter than five you could check the second, and so on?

bassosimone: For example, `kernel.org` title is `The Linux Kernel Archives`

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,546 @@
...
+            return True
+
+        control_addrs = set(self.control['dns']['addrs'])
+        experiment_addrs = set(experiment_dns_answers)
+
+        if control_addrs == experiment_addrs:
+            return True
+
+        for experiment_addr in experiment_addrs:
+            if is_public_ipv4_address(experiment_addr) is False:
```
bassosimone: `if not is_public...` ?

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -0,0 +1,546 @@
...
+                           control_addrs))
+
+        # Remove the instance of AS0 when we fail to find the ASN
+        control_asns.discard('AS0')
+        experiment_asns.discard('AS0')
+
+        if len(control_asns.intersection(experiment_asns)) > 0:
+            return True
+
+        return False
```
bassosimone: The return statement here and above could be simplified as `return len(...) > 0`

bassosimone:
Update: completed a second reading of the diff. What I have seen looks good. Testing tomorrow.

ooni/utils/net.py
hellais: yup.

ooni/templates/httpt.py
hellais: Yes, it's actually not used in this module at all. It's probably a leftover from when I moved stuff into common. Will fix.

ooni/backend_client.py
hellais: True.

ooni/backend_client.py
hellais: +1.

ooni/nettests/blocking/web_connectivity.py
hellais: yeah, sure.

ooni/nettests/blocking/web_connectivity.py
hellais: Yes, true. This is a good point. I will implement something like that.

bassosimone:
@hellais The following diff needs to be applied to also install stuff in `ooni/common`:

```diff
diff --git a/setup.py b/setup.py
index da0b967..3bfc5cd 100644
--- a/setup.py
+++ b/setup.py
@@ -171,6 +171,7 @@ data_files = []
 packages = [
     'ooni',
     'ooni.api',
+    'ooni.common',
     'ooni.deckgen',
     'ooni.deckgen.processors',
     'ooni.kit',
```

ooni/nettests/blocking/web_connectivity.py
hellais: I think for this I would like to leave it explicit like so, as to make it clearer what are the all the possible outcomes of this comparison.

hellais:
Oh, right I forgot to update the setup.py.

bassosimone:
@hellais, one more thing: I think we should log at level msg to tell the user when we're performing the control request with the backend as well:

```diff
diff --git a/ooni/nettests/blocking/web_connectivity.py b/ooni/nettests/blocking/web_connectivity.py
index 655bf76..0882e16 100644
--- a/ooni/nettests/blocking/web_connectivity.py
+++ b/ooni/nettests/blocking/web_connectivity.py
@@ -230,6 +230,7 @@ class WebConnectivityTest(httpt.HTTPTest, dnst.DNSTest):
 
     @defer.inlineCallbacks
     def control_request(self, sockets):
+        log.msg("* performing control request with backend")
         self.control = yield self.web_connectivity_client.control(
             http_request=self.input,
             tcp_connect=sockets
```

ooni/nettests/blocking/web_connectivity.py
bassosimone: ok

bassosimone:
So, I have run some tests with specific URLs (some of which were taken from [citizenlab/test-lists](https://github.com/citizenlab/test-lists) and `HEAD` equal to 3aa17317012a (with the two patches pasted above applied) from the `probe` Vagrant box. The command I run in each case was:

```
ooniprobe -b http://riggrf3c33dnnzuu.onion blocking/web_connectivity --url $URL
```

Here are some interesting things I noticed. (@hellais, I assume that, since you set up the backend I used, you could also access reports given the IDs, hence in the following I am not pasting the full `.yamloo` files but just the bare minimum required to make my points.) I am numbering my test cases so that later it would hopefully be easier to comment on them.

# 1. Popcorntime

```yaml
---
options: [--url, popcorntime.se]
report_id: 20160530T081204Z_AS3269_FdHDIfcOZgubo3j1CoD4kDZA0XR1Rrv1min2d4HVLkJbTliWzB
...
---
control:
  dns:
    addrs: [195.74.38.149]
    failure: null
  http_request:
    body_length: -1
    failure: generic_timeout_error
    headers: {}
    status_code: -1
  tcp_connect:
    195.74.38.149:443: {failure: unknown_error, status: false}
input: https://popcorntime.se
```

Here the `tcp_connect` failure in the control is `unknown_error`. It is perhaps one of those cases where adding more specific errors to the backend would help. By the way, I tried to reproduce the error for such input with my own backend, but in my case I got `connection_refused_error` as error.

Also, an perhaps off topic with respect to this pull request, it is not totally clear to me why a `tcp_connect` could fail with `connection_refused_error` and a `http_request` could at the same time fail with a `generic_timeout_error`; shouldn't the `http_request` code notice that the connection was refused?

# 2. Kernel

```
---
options: [--url, '']
report_id: 20160530T081744Z_AS3269_W3raDgH3yspV6JgXQ3bH48STmzEs2OTmdkmQOURMPc4LBW5J7t
...
---
control:
  http_request:
    headers: {Accept-Ranges: bytes, Content-Type: text/html; charset=UTF-8, Date: 'Mon,
        30 May 2016 08:24:05 GMT', ETag: '"570f-53406cfc81e00"', Last-Modified: 'Mon,
        30 May 2016 03:25:44 GMT', Server: nginx, Strict-Transport-Security: max-age=15768000,
      X-Frame-Options: DENY, content-encoding: ''}
headers_match: false
input: https://kernel.org/
requests:
- failure: null
  response:
    headers: {Accept-Ranges: bytes, Content-Type: text/html; charset=UTF-8, Date: 'Mon,
        30 May 2016 08:24:04 GMT', ETag: '"570f-53406cfc81e00"', Last-Modified: 'Mon,
        30 May 2016 03:25:44 GMT', Server: nginx, Strict-Transport-Security: max-age=15768000,
      X-Frame-Options: DENY, content-encoding: ''}
retries: 1
```

Here the following things are not totally clear to me:

a. why was the URL empty in `options`?

b. it seems to me that the only header that changed is `Date`, which is supposed to be the date when the server emitted the response and is supposed to vary with time, hence I think here the `headers_match` should be `true` as opposed to `false`

c. not totally clear to me the semantic of `retries` which I have seen equal to `1` both when no retries was made and when they were made (this is perhaps something I should have learned reading the code but apparently I failed to do so)

# 3. KernelTrap

```yaml
---
options: [--url, '']
report_id: 20160530T082127Z_AS3269_KpGZ10DxLJrEKeDfBRLyPBMTf3b3p10ebTIUpSrimmZZTNdoBz
test_helpers:
  backend: {address: 'httpo://ckjj3ra6456muu7o.onion', type: onion}
...
---
input: https://kerneltrap.org/
requests:
- failure: generic_timeout_error
- failure: generic_timeout_error
- failure: generic_timeout_error
retries: 1
```

Here again I do not have any URL in `options`. Also, now that I notice it, I find it striking that the `backend` onion address is different from the one specified on command line. Finally, this is to show that I have `retries` equal to `1` also with three failures in making requests.

# 4. PopPurchase

```yaml
---
options: [--url, www.popurchase.com]
report_id: 20160530T083249Z_AS3269_at4IIbbONwVsNDUIoFTCVJTsSG2m2i4Ws2exHe64XWVLxQXpkc
...
---
body_length_match: true
body_proportion: 1.0
control:
  dns:
    addrs: [165.160.15.20, 165.160.13.20]
    failure: null
  http_request:
    body_length: 23903
    failure: null
    headers: {Age: '3', Content-Type: text/html; charset=UTF-8, Date: 'Mon, 30 May
        2016 08:43:06 GMT', Link: '<http://gbcinternetenforcement.net/?p=6126>; rel=shortlink',
      Server: nginx/1.2.4, Via: 1.1 varnish, X-Pingback: 'http://gbcinternetenforcement.net/xmlrpc.php',
      X-Powered-By: PHP/5.3.6-13ubuntu3.10, X-Varnish: 912428719 912428652, content-encoding: ''}
input: http://www.popurchase.com
requests:
- failure: null
  request:
    method: GET
    tor: &id001 {exit_ip: null, exit_name: null, is_tor: false}
    url: http://gbcinternetenforcement.net/14-9752
```

So, in this case what is interesting is that:

a. the `URL` appears again in `options` (now that I notice it, it is somewhat parsed because I passed a URL with a leading `http://` explicitly as my `history` shows)

```
   36  ooniprobe -b http://riggrf3c33dnnzuu.onion blocking/web_connectivity --url http://www.popurchase.com
```

b. the final request that I show is after a redirect (omitted); what is interesting in this case (try to visit the URL yourself) is that the website is seized (i.e. this is a sort of "blocking page") but we are not able to notice this because both the measurement and the control end up on the seized site; I wonder whether we should treat this specific case specially (strictly speaking I agree that the experiment succeeded but perhaps it would be interesting that we land on a domain seized page).

hellais:
Regarding the last comment:

> 1. Popcorntime

I did some other tests and I see that it gives this unknown error.

Looking into it and adding some extra debugging I can say the error is actually 
```
twisted.internet.error.ConnectError:
An error occurred while connecting: 113: No route to host.
```

I ran it multiple times and I always got unknown error in both the http response and the tcp connect failure. I am not sure why you are seeing these inconsistencies.

I will add these other mappings to the backend.

> 2. Kernel

a. So the issue with the `options` field is that we do some sanitisation to avoid leaking user paths (https://github.com/TheTorProject/ooni-probe/blob/754ce68db5e11e0d9b6df7fbc5e7b898489af061/ooni/utils/__init__.py#L140) since we don't really care to much that the options are accurate (i.e. the information from the option is then also part of the report). I think for the moment it's ok to leave this as is.

b. Yes, currently we check to see only if there are common headers amongst the ones that are "rare" (not inside of https://github.com/TheTorProject/ooni-probe/blob/feature/web_connectivity/ooni/utils/net.py#L51). This means that if a server replies with **only** common headers we consider the headers to not match. I am going to change this to make it so if **all** headers match even if they are all common ones we say headers match (Note: by match we mean the header keys match, not the values). While keep the logic of finding common "rare" headers for the rest. In my testing I have seen this reduces some false positives.


c. since retries can also be set by the user. They can set it to something different than 1 and we want to record that number inside of the report.


> 4.  PopPurchase

Regarding detection of siezed websites, this is kind of tricky to do and also unclear if it's something we should be doing as part of ooniprobe. We are generally interested in differences in accessibility that is **specific** to the vantage point of the probe, so in theory siezing of a website would not fit under this class.

I think it's worthwhile nonetheless opening a ticket about this to discuss it further and consider it including in the future.

hellais:
https://github.com/TheTorProject/ooni-backend/pull/85

bassosimone:
> [...]
> c. since retries can also be set by the user. They can set it to something different than 1 and we want to record that number inside of the report.

This point is not clear to me (and ack for everything else). So, this number is the default number of retries of what? To clarify: what I find strange is this number being equal to `1` in the report and three retries being made when the server is non reachable.

hellais:
> This point is not clear to me (and ack for everything else). So, this number is the default number of retries of what? To clarify: what I find strange is this number being equal to 1 in the report and three retries being made when the server is non reachable.

Yes so this is the number of retries for the HTTP request experiment measurement. There are also some retries that are associated in general to the whole test run and these are handled by ooniprobe itself. In the case above you triggered a `generic_timeout_error` that happens when a test runs for more than the timeout set for it to complete (in the case of the web_connectivity test it's more than 240 seconds.

bassosimone:
> In the case above you triggered a generic_timeout_error that happens when a test runs for more than the timeout set for it to complete

Ah, got it. I thought the timeout was on the connect and not on the whole test. (Perhaps, it would make sense to have an error for that in the future?)

bassosimone:
I have reviewed the results of a deck that I run (and interrupted after half an hour) this morning without seeing stack traces or other serious errors. I think we can merge at your earliest convenience, @hellais!

PR #462, opened 2016-03-11
Remove obsolete install method on Debian/Ubuntu

anadahz:
Currently the ooniprobe packages offered in Debian and Ubuntu are old,
unavailable in some releases or very old version. Removing obsolete
install instructions until the ooniprobe packages in the repositories
are available and up-to-date


hellais:
Sounds good.

PR #470, opened 2016-03-30
fix the dependency on pyasn1

### LINKED ISSUE #469: ooniprobe depends on twisted-16 which depends on pyasn1>=0.1.8
Installed /usr/local/lib/python2.7/dist-packages/Twisted-16.0.0-py2.7-linux-x86_64.egg

error: pyasn1 0.1.7 is installed but pyasn1>=0.1.8 is required by set(['pyasn1-modules'])

----------------------------------------
Cleaning up...
Command /usr/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-wOaTLz/ooniprobe/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-hVf_ku-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip-build-wOaTLz/ooniprobe
  File "/usr/bin/pip", line 9, in <module>
    load_entry_point('pip==1.5.6', 'console_scripts', 'pip')()
  File "/usr/lib/python2.7/dist-packages/pip/__init__.py", line 248, in main
    return command.main(cmd_args)
  File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 161, in main
    text = '\n'.join(complete_log)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 29: ordinal not in range(128)

vmon:
close #469

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ scapy-real>=2.2.0-dev
 pypcap>=1.1
 service-identity
 pydumbnet
-pyasn1
+pyasn1>=0.1.8
 pyasn1-modules
 characteristic
 zope.interface


hellais:
I am thinking that since ooniprobe itself doesn't depend directly on pyasn1 and twisted-16 should have pyasn1 with the correct version in it's dependencies, it's perhaps wiser to just remove the dependency from **our** requirements.txt and make it be included when resolving the dependencies of twisted?

hellais:
Ok so I have looked more into this and it appears you are right. We need to manually specify this as the required pyasn version, since if a previous version of pyasn is found on the system then it will not attempt to upgrade the version when pulling in the sub-depedencies 😵

PR #475, opened 2016-04-16
Enable test with default options

anadahz:


anadahz:
We should fix the tests before merging this.

hellais:
In order to get this to work a working backend is needed. Probably it's preferable, though, to use the bouncer routines to look it up.

hellais:
Issue moved to [ooni/probe-legacy #9](https://github.com/ooni/probe-legacy/issues/9) via [**ZenHub**](https://www.zenhub.com/)

PR #485, opened 2016-04-28
Check to ensure response_body is either a str or unicode before sanit…

hellais:
…ising it

Failure to do so will lead to sanitisation issues when the response body is binary.


bassosimone:
The diff looks good to me!

PR #492, opened 2016-05-03
Add Whatsapp deck and helper script for CIDR

anadahz:
A Whatsapp ooni-probe deck with a shell script deck helper to grab and
process Whatsapp's CIDR IPs. The Deck runs the following ooniprobe tests:
- http_requests test on Whatsapp's HTTP/S URLs
- dns_consistency test on Whatsapp's domain/subdomains
- http_hosts tests on all oficial Whatapps CIDR IPs connecting on all
  know ports (80, 443, 5222, 5223, 5228, 5060, 5060, 8080).
The deck helper script grabs the official CIDR IP list that Whatsapp
uses to connect, extracts all the IPv4 addresses and converts the CIDR
notations to IPs and finally splits them into list files per Whatsapp
port.


hellais:
This is a very good idea in general.

However there are a couple of problems with how this is implemented:

* The script to download the list of IPs used by whatsapp should not be a shell scripts, but should rather be part of ooniresources.

* The http_host test is not the right test to ensure connectivity to an IP address. As the name implies it's looking for filtering based on hostname parsing by attempting to connect to a control backend and setting the `Hostname` HTTP header.

hellais:
I propose the following changes to this deck. I pushed them to a branch named `feature/whatsapp-deck-fix`.

hellais:
What should we do about this branch? Should we close this PR and merge `feature/whatsapp-deck-fix`?

anadahz:
Yes closing this PR let's merge `feature/whatsapp-deck-fix`.

PR #506, opened 2016-05-18
The version number in master should always be greater than the latest…

hellais:
… release

This is to allow us to distinguish reports that are run from ooniprobe versions
cloned from master and those run from packages.

### DIFF
--- a/ooni/__init__.py
+++ b/ooni/__init__.py
@@ -1,7 +1,7 @@
 # -*- encoding: utf-8 -*-
 
 __author__ = "Open Observatory of Network Interference"
-__version__ = "1.4.2"
+__version__ = "1.5.0-master"
 
 __all__ = ['config', 'inputunit', 'kit',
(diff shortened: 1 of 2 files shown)


bassosimone:
Looks good to me! Will merge as soon as automatic checks complete!

bassosimone:
@hellais, I don't immediately understand how, but [this diff is breaking the build](https://travis-ci.org/TheTorProject/ooni-probe/jobs/131115723#L1783).

bassosimone:
Ah, right! Guess I've figured it out, probably the issue is that it tries to download ooni resources from github using a version number that does not exist yet.

hellais:
Yes you are right.

I came up with this solution that is a bit dirty and requires a slight bit more of hassle when doing a new release, but I can't think of something better for the moment.

bassosimone:
> I came up with this solution that is a bit dirty and requires a slight bit more of hassle when doing a new release, but I can't think of something better for the moment.

Yeah, for now it would do! Merging!

PR #520, opened 2016-06-02
Fix/remove meek google test

anadahz:


anadahz:
I added some unrelated manpages doc fixes to this PR my bad cleaning up.

Issue #503, opened 2016-05-16
Adding bisection logic to inputProcessor


seamustuohy:
I have mocked up a few nearly working version of keyword list bisection  and have been playing with placing the bisection logic into the inputProcessor. I did this to separate the keyword testing components from the keyword bisection components. This made the mocked up code is considerable cleaner.

In order to make it functional the individual tests need to be able to pass data to the main test_case.inputs generator that seeds each measurement.

​https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettest.py#L528-L538

I see this being done in one of two ways.

The first is to leverage an outside data-store that the inputProcessor is pulling from, and the test is populating with results. This adds the addition of new services into the stack for a single test, which I don't see as a useful addition. 

The second option is to use the send() function to push values into the generator. This would be a clean way to provide feedback to the test_case.inputs generator that it could use to bisect the blocked input and provide them on a later iteration. I currently cannot use this method because the test instance passed to each measurement is a new test_case object, and therefore does not have access to the test_case.inputs object that is generating inputs. 

I would be happy to take a pass at passing reference to the test_case.inputs object to the test to get a working prototype working if that is the right path to head down. I wanted to ask in here to see if there might be a more appropriate way. I don't want to head down this path if there are reasons why this would never be merged into the core codebase.

Here is my current bisection functionality mock-up for the interested:
​https://github.com/elationfoundation/ooni-probe/blob/feature/bisect_inputs/ooni/nettests/experimental/bisect_send.py

hellais:
Thanks for putting to the test the OONI test writing API in a very interesting way!

First let me try to understand if I properly understand the problem.

Basically you want to keep some state across different test_cases and have the input of the test_case be set based on this "accumulated" state.
As you rightfully point out `inputs` is currently being reset inside of the `_setUp` method of the `NetTestCase`.
I have looked a bit at the code in question and I don't see a particular reason for us to need to reset the value of inputs. So I think you can even avoid setting `self.inputGenerator  = self.inputs` and modifying the base `NetTest` class to remove the line `self.inputs = None` here: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettest.py#L709.

I like the approach of sending values to the `self.inputs` generator and I think this is indeed the cleanest solution so I would take this path.

Looking at your code you should ensure that you yield the packet also a second time because otherwise you will end up skipping some values.
Quoting from (https://docs.python.org/2/reference/expressions.html#generator.send)

[...]

PR #523, opened 2016-06-05
Make inputs shared across all NetTestCase measurements

elationfoundation:
This pull request makes the inputs iterator shared by all the measurement tests within a NetTestCase. Per issue #503 , In order to make bisection style testing functional the individual tests need to be able to pass data to the main test_case.inputs generator that seeds each measurement.  This provides opportunities for each measurement test in a NetTestCase to communicate with the iterator to do things such as adding additional values to be passed to measurement tests as a later input.

In the files changed you will find the very-small code update required for this feature, unit-tests that ensure proper functionality, and updates to the user documentation to guide others in effectively taking advantage of it.

### DIFF
--- a/docs/source/writing_tests.rst
+++ b/docs/source/writing_tests.rst
@@ -57,13 +57,70 @@ this::
             yield x.strip()
         fp.close()
 
-For example, if you wanted to modify inputProcessor to read enteries from a CSV file, you could use::
-            
+For example, if you wanted to modify inputProcessor to read entries from a CSV file, you could use::
+
     def inputProcessor(self, filename):
         with open(filename) as csvFile:
             reader = DictReader(csvFile)
             for entry in reader:
                 yield entry
+
+
+The ``inputs`` iterator is unique per ``NetTestCase`` and shared by all its measurement tests. This provides opportunities for each measurement test in a ``NetTestCase`` to communicate with the iterator to do things such as adding additional values to be passed to measurement tests as a later ``input``.
+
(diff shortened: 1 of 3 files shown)


hellais:
Good stuff! Thanks for hacking on this. I will review it and look into merging it over the next week.

elationfoundation:
Oh, I almost forgot [here](https://github.com/elationfoundation/ooni-probe/blob/example/writing_tests_gen/ooni/nettests/experimental/test_send.py) is the example code from writing_test.rst made into an actual nettest. I wanted to make sure that the code example actually worked.  Hope it saves you a few minutes when you are reviewing the documentation updates.

hellais:
So I have been reviewing this and I think the logic is sound. However due to the sending of new inputs to the generator, with a large amount of inputs (> 60) we run into a maximum recursion problem.

This can be reproduced simply by running something like:

```
from ooni import nettest
from twisted.internet import defer

class UrlGeneratorWithSend(object):
    def __init__(self):
        """Create initial list and set generator state."""
        self.urls = []
        for x in range(60):
            self.urls.append("input-{0}".format(x))
        self.current = 0

    def __iter__(self):
        return self

    def __next__(self):
        try:
            cur = self.urls[self.current]
            self.current += 1
            return cur
        except IndexError:
            raise StopIteration

    # Python 2 & 3 generator compatibility
    next = __next__

    def send(self, returned):
        """Appends a value to self.urls when activated"""
        if returned is not None:
            print("Value {0} sent to generator".format(returned))
            self.urls.append(returned)


class TestUrlList(nettest.NetTestCase):

    # Adding custom generator here
    inputs = UrlGeneratorWithSend()

    def postProcessor(self, measurements):
        """If any HTTPS url's are passed send back an HTTP url."""
        if not self.input.startswith("double-"):
            self.inputs.send("double-{0}".format(self.input))
        return self.report

    def test_url(self):
        self.report['tested'] = [self.input]
        return defer.succeed(1)
```

Digging into the matter I have figured out that this was a concurrency issue with the scheduling algorithm that even had a very nasty comment: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/managers.py#L64

By doing some locking on the tasks generator this get's solved nicely and it ends up working as expected.

I am going to proceed in merging this with my patch applied on top of this.

Thanks a bunch for this.

Issue #483, opened 2016-04-27
Make it possible to run tests without specifying the test type


hellais:
It would be nice it were possible to run tests without having to specify the category they belong to.

For example to run the `tcp_connect` test instead of:

```
ooniprobe blocking/tcp_connect
```

just do

```
ooniprobe tcp_connect
```

hellais:
Made it into master!

PR #532, opened 2016-06-10
Add support for specifying the test name without the category prefix.

hellais:
This implements: https://github.com/TheTorProject/ooni-probe/issues/483

### DIFF
--- a/ooni/deck.py
+++ b/ooni/deck.py
@@ -73,6 +73,9 @@ def nettest_to_path(path, allow_arbitrary_paths=False):
     """
     Takes as input either a path or a nettest name.
 
+    The nettest name may either be prefixed by the category of the nettest (
+    blocking, experimental, manipulation or third_party) or not.
+
     Args:
 
         allow_arbitrary_paths:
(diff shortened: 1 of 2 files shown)


bassosimone:
I am currently testing this!

ooni/deck.py
```diff
@@ -85,11 +88,30 @@ def nettest_to_path(path, allow_arbitrary_paths=False):
...
         return path
 
-    fp = FilePath(config.nettest_directory).preauthChild(path + '.py')
-    if fp.exists():
-        return fp.path
-    else:
+    p = path.split("/")
+    test_name = p[0]
+    if len(p) > 1:
+        test_categories = [p[0]]
```
bassosimone: This assignment is redundant because below you assign unconditionally to `test_categories`

ooni/deck.py
```diff
@@ -85,11 +88,30 @@ def nettest_to_path(path, allow_arbitrary_paths=False):
     if allow_arbitrary_paths and os.path.exists(path):
         return path
 
-    fp = FilePath(config.nettest_directory).preauthChild(path + '.py')
-    if fp.exists():
-        return fp.path
-    else:
+    p = path.split("/")
```
bassosimone: Would it make sense here to use `path.rsplit("/", 1)`?

bassosimone:
- [ ] testing
- [ ] diff review
- [ ] fixes
- [ ] merging!

ooni/deck.py
hellais: Fair point. I guess we could do an `rsplit` and get the last item as the test_name and ignore the test_category section.

bassosimone:
Tested again after d78e7d4, it works! Merging!

PR #549, opened 2016-06-17
Fix/man page

anadahz:
This fixes https://trac.torproject.org/projects/tor/ticket/17319

cc: @willscott

### DIFF
--- a/data/ooniprobe.1
+++ b/data/ooniprobe.1
@@ -1,7 +1,7 @@
 .
-.TH "ooniprobe" "1" "April 26, 2016" "1.4.0" "ooniprobe"
+.TH "OONIPROBE" "1" "April 26, 2016" "1.4.0" "ooniprobe"
 .SH NAME
-ooniprobe - a network censorship measurement tool.
+ooniprobe \- a network censorship measurement tool.
 .
 .nr rst2man-indent-level 0
 .
@@ -57,8 +57,8 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 
 .SH DESCRIPTION
 .sp
-ooniprobe is tool for performing internet censorship measurements. Our goal is
-to achieve a common data format and set of methodologies for conducting
+\fBooniprobe\fP is the tool for performing internet censorship measurements. Our
+goal is to achieve a common data format and set of methodologies for conducting
 censorship related research.
 
 .SH OPTIONS
@@ -137,7 +137,7 @@ tampering with the internet traffic between the probe and a remote test helper
 backend. As such they usually require the selection of a oonib backend
 component for running the test.
 .sp
-\fBContent Blocking\fP are aimed at enumerating the kind of content that is
+\fBContent Blocking\fP is aimed at enumerating the kind of content that is
 blocked from the probes network point of view. As such they usually require to
 have specified an input list for running the test.
 .SS Threat Model
@@ -171,7 +171,7 @@ unrestricted
 .sp
 \fINote\fP that while 2) => 1) it is not true that 1) => 2) as the identification of
 such actors does not necessarily have to happen in real time.
-While our intention is to minimize the risk of users running OONI probe to be
+While our intention is to minimize the risk of users running ooniprobe to be
 identified, this comes with a tradeoff in accuracy. It is therefore necessary in
 certain tests to trade\-off fingerprintability in favour of tests accuracy.
 .sp
--- a/data/ooniprobe.1
+++ b/data/ooniprobe.1
@@ -227,16 +227,6 @@ You may then run a deck by using the command line option \fI\-i\fP:
 .sp
 As root:
 .sp
-\fIooniprobe \-i /usr/share/ooni/decks/mlab.deck\fP
-.sp
-Or as a user:
-.sp
-\fIooniprobe \-i /usr/share/ooni/decks/mlab_no_root.deck\fP
-.sp
-Or:
-.sp
-As root:
-.sp
 \fIooniprobe \-i /usr/share/ooni/decks/complete.deck\fP
 .sp
 Or as a user:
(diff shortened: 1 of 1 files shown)


data/ooniprobe.1
```diff
@@ -119,6 +121,9 @@ Useful when debugging freezes or locks in complex code.
 .TP
 .BR \-\-version
 Display the ooniprobe version and exit.
+.TP
+.BR \-\^Q " or " \-\-queue
+AMQP Queue URL amqp://user:pass@host:port/vhost/queue
```
willscott: What is an 'AMQP'? This needs a better description.

data/ooniprobe.1
```diff
@@ -119,6 +121,9 @@ Useful when debugging freezes or locks in complex code.
...
 
 .SH OONIPROBE
 .sp
```
willscott: on line 130: "researches" -> "researchers"

willscott:
These look good, but there are remaining nits from the ticket in trac that aren't addressed by this, like reference to missing mlab test decks among others. Is this meant to be a combined pull to fix everything?

data/ooniprobe.1
anadahz: In sort this is used to make ooniprobe work with blocked: https://github.com/openrightsgroup/cmp-issues/issues/78
Not sure how we can describe this in a way that makes senses to the user.

data/ooniprobe.1
anadahz: Thx fixed.

anadahz:
@willscott Do you think that mlab test decks are relevant to the ooniprobe user?

anadahz:
It will be nice to have an auto generated man page with the actual ooniprobe version and date.

willscott:
I think they aren't relevant, which is why having them referenced in the man page is potentially confusing.

The queue thing needs at least sufficient documentation for someone technical to figure out what's going on. even that github bug leaves it pretty unclear what's going on here. Is this some alternative way to get test input (that's stored in a rabbitmq server somewhere?)

Saying, 'there's this option that causes a set of acronyms to happen' seems like it can only increase confusion.

anadahz:
Removed the mlab deck references.
I agree the AMQP option introduces confusion. 
I'm not aware of a way to get a test input.
Perhaps @dantheta can provide as with some useful test input?

dantheta:
I don't think there is a way to get test input for the AMQP routines - it's intended to point at an AMQP (Rabbit) server.  However, the AMQP options are only useful on the bin/ooniprobequeue command (which uses a variant director), not bin/ooniprobe; I guess the command line options appear for ooniprobe because the command-line options are in shared code.

It might be possible to wrap the parseOptions function so that it uses an extended Options class for the DaemonDirector.  That way the --amqp option wouldn't appear when someone was running bin/ooniprobe --help, only when they were running bin/ooniprobequeue.

willscott:
an initial set of additional changes I'll propose: https://github.com/willscott/ooni-probe/commit/aab516cc433c2cc53bbe0f22d3379f962cc92f69

anadahz:
@willscott apart from the queue comments (https://github.com/willscott/ooni-probe/commit/aab516cc433c2cc53bbe0f22d3379f962cc92f69#commitcomment-18060250) looks good to me.

willscott:
Re-added AMQP help line to my branch https://github.com/TheTorProject/ooni-probe/compare/fix/man-page...willscott:fix/man-page

And filed bug to do it the right way as suggested https://github.com/TheTorProject/ooni-probe/issues/560

Merge/squash my branch into this pull request?

hellais:
@dantheta actually since ooniprobe 1.4.0 (https://github.com/TheTorProject/ooni-probe/commit/1192ca0d0fd7c7b4aa5cb593fe711eb175d5c880 specifically) there is no longer the need to use the specialised ooniprobequeue command to use the messagequeue, but it will invoke that code path upon launching it with the `--queue` option.

You should probably keep this in mind when you update to a later version of ooniprobe for ORG.

dantheta:
Ah, that's really cool.  Thanks!

willscott:
Lets also try to improve the documentation of the -p option, which is unclear (https://trac.torproject.org/projects/tor/ticket/11988)

hellais:
I believe this is now superseeded by the manpage docs being generated from markdown files. I believe all the feedback in here should be integrated in there, but please double check to be sure.

hellais:
Can I close this?

anadahz:
There is significant amount of documentation missing from the markdown docs that should either be added to the docs itself or referenced to the manpages as links or external files (eg: in /usr/share/doc).

The documentation topics that are not covered by the current [markdown manual pages](https://github.com/TheTorProject/ooni-probe/tree/master/docs/source/manual) are:
- Default collector, bouncer addresses
- Test enumeration and description
- Default paths for reports, decks and config files
- OONI terms explanation (tests, decks, test helper,...)
- Configuring ooniprobe
- Running decks
- Running net tests
- Optional packages (obfs4proxy)
- Bridges and obfsproxy bridges
- Third party tests and dependencies

hellais:
> There is significant amount of documentation missing from the markdown docs that should either be added to the docs itself or referenced to the manpages as links or external files (eg: in /usr/share/doc).

Sounds good. Can you add these to the markdown files as part of a new PR? I also don't think all of these need to necessarily be present inside of the manpage as they are documented elsewhere.

anadahz:
@hellais Sure, though if this issue is closed it will be most probably forgotten.

> I also don't think all of these need to necessarily be present inside of the manpage as they are documented elsewhere.

Which places are you referring to?
It makes sense to add a URL in the man pages pointing to the documentation already mentioned elsewhere.

hellais:
> Which places are you referring to?

The README has most of these things.

> It makes sense to add a URL in the man pages pointing to the documentation already mentioned elsewhere.

Sure go for it!

PR #550, opened 2016-06-18
Make sure oonireport receives correct args

bassosimone:
> *Tu quoue, argv[0], fili mi!*

Closes #457 reported by @anadahz.

Tested locally (this time!) by running:

```
export PYTHONPATH=.
./bin/oonireport status
./bin/oonireport --version
./bin/oonireport upload
```

All commands now seem to work as they should.

### DIFF
--- a/ooni/report/cli.py
+++ b/ooni/report/cli.py
@@ -60,7 +60,7 @@ def tor_check():
         sys.exit(1)
 
 
-def run(args=sys.argv):
+def run(args=sys.argv[1:]):
     options = Options()
     try:
         options.parseOptions(args)


hellais:
Oh I missed this before writing and opening https://github.com/TheTorProject/ooni-probe/pull/552.

Though I think passing as a default argument `None` is more robust than doing `sys.argv[1:]` as the later can potentially lead to an index out of range error if `len(sys.argv) < 2`.

Passing argument None to parseOptions on the other hand will lead to reading `sys.argv` as part of `twisted.python.usage`

hellais:
Ignore my previous comment, actually string slicing will always return at least an empty list even if the starting index is > than the length of the array.

I think this is actually good to go.

PR #551, opened 2016-06-19
Update README to use deb.tpo instead of irl's private PPA

irl:


irl:
Someone might want to fix my terrible attempt at rst formatting, but the note there I think covers it well.

hellais:
Fixed the markup by removing the italics and labelling it as `Note:`. Thanks for this!

PR #559, opened 2016-06-28
Make the imports of the CollectorClient conditional

hellais:
This brings down the runtime of the ooniprobe --version command from
1.302 seconds to 0.464 seconds (280% improvement).


hellais:
As this really doesn't add any more logic I am just going to merge this.

PR #561, opened 2016-06-29
Write the deck file to always the same filename deck/

hellais:
- Add some basic unittests for oonideckgen


hellais:
We have since deprecated ooni-deckgen. This is no longer relevant.

PR #562, opened 2016-06-30
Feature/1.6.0 release

hellais:
This shall be merged once we have deployed the production bouncer and backends (https://github.com/TheTorProject/ooni-sysadmin/issues/61)

Note: this is going to fail because the resources with the 1.6.0 version number cannot exist until we upload them.

### DIFF
--- a/ooni/__init__.py
+++ b/ooni/__init__.py
@@ -1,10 +1,10 @@
 # -*- encoding: utf-8 -*-
 
 __author__ = "Open Observatory of Network Interference"
-__version__ = "1.6.0rc1"
+__version__ = "1.6.0"
 # This is the version number of resources to be downloaded
 # when a release is made it should be aligned to __version__
-__resources_version__ = "1.5.1"
+__resources_version__ = "1.6.0"
 
 __all__ = [
     'common',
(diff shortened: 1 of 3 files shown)


bassosimone:
> Note: this is going to fail because the resources with the 1.6.0 version number cannot exist until we upload them.

@hellais this is a chicken-and-egg problem I also have with MK, where the pre-compiled dependencies could not be uploaded until there is a release. For now, I have managed to solve the problem by putting the dependencies into a previous release, which sucks. It would probably make sense to publish those files in a place that is not bound to a specific release, wouldn't it?

bassosimone:
@hellais is it correct to say that this pull request cherry-picks patches from other, opened PRs?

bassosimone:
> @hellais is it correct to say that this pull request cherry-picks patches from other, opened PRs?

Never mind, I cannot find an open pull request with that content inside. Yet, reading the diff I have had a sensation of *deja vu*. Is me reviewing this pull request a fixed point in time?

hellais:
Oh my, you are right. I think it's no longer properly rebased against master.

hellais:
This now landed in master 🎈

PR #567, opened 2016-07-12
Feature/webui2

hellais:
I will be using this branch to keep track of the work on the ooniprobe web-ui.

(this superseeds #442 that superseeded #338)

I am quite satisfied with the overall code architecture based on angular components and I wrote a bit of documentation in the Readme that explains the rational behind it.

I think we should be giving higher priority to the development of the views for the actual measurement results since that is something that is also needed for net-probe (see https://github.com/measurement-kit/measurement-kit-app-ios/issues/33)

cc @willscott @lorenzoPrimi


hellais:
yeah I agree that if we want this to be shared across repositories it's probably best to split it out into another repository. I will go ahead in creating ooni-wui that shall contain all the web UI related assets and files. We can then include as part of ooniprobe just a built version of the web UI based on that.

hellais:
Ok so I split this up into it's own repository and it now lives here: https://github.com/TheTorProject/ooni-wui.

I also implemented a proof of concept of the measurement visualiser to be used by net-probe.

@lorenzoPrimi if you build what is inside of https://github.com/TheTorProject/ooni-wui with `gulp build --mobile --fixtures` you will see how it works. You can then use the HTML and JS inside of the `dist/build/` directory and remove the line that is including the test fixtures and try it out with what I suggested in https://github.com/measurement-kit/measurement-kit-app-ios/issues/33 and you should be able to see something.

lorenzoPrimi:
@hellais I implemented this in iOS but I can't see anything in the webview, I only see a "developer info" button. I attach some MeasurementJSON got from the app.

[various_js.zip](https://github.com/TheTorProject/ooni-probe/files/379778/various_js.zip)

hellais:
@lorenzoPrimi Looking at the JSON in the attached zip the file `httpinvalidrequestline.js` is formatted properly, however `tcpconnect.js` and `dnsinjection.js` are not, since they contain a series of JSON documents separated by newlines. In order to have them render correctly you need to place as the return value of the `get()` function a single document (the one you want to display).

lorenzoPrimi:
How does the user select which result to display? In the app there is only one "Result" button, as that json file is associated to a single test

hellais:
Moving this to #581.

PR #570, opened 2016-07-19
Fix #569

### LINKED ISSUE #569: Failure to run ooniprobe test deck without submiting reports to collector
`ooniprobe version: 1.6.0`

`$ ooniprobe -v -n -i /home/user/ooni-probe/deck-XX/default-user.deck`
```
Log opened.
Not reporting using a collector
[D] Checking if backend is present
[D] Checking if backend is present
[D] Checking if backend is present
[D] Checking if dns-discovery is present
[D] Tor is not running. Skipping IP lookup via Tor.
Looking up your IP address via torproject
Looking up collector and test helpers with http://nkvphnp3p6agi5qq.onion
Unhandled Error
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1504, in run_script
    exec(code, namespace, namespace)
  File "/home/user/ooni-probe/ooni-env/lib/python2.7/site-packages/ooniprobe-1.6.0-py2.7.egg/EGG-INFO/scripts/ooniprobe", line 23, in <module>
    reactor.run()
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/base.py", line 1194, in run
    self.mainLoop()
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/base.py", line 1203, in mainLoop
    self.runUntilCurrent()
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/base.py", line 825, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 419, in resolveAddress
    self._setRealAddress(self.addr)
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 434, in _setRealAddress
    self.doConnect()
  File "/home/user/ooni-probe/ooni-env/local/lib/python2.7/site-packages/twisted/internet/tcp.py", line 563, in doConnect
    connectResult = self.socket.connect_ex(self.realAddress)
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
exceptions.TypeError: an integer is required
```

hellais:
"foo" == u"foo" is True, but "foo" is u"foo" is False.

Oh python, what you make me do...

### DIFF
--- a/ooni/deck.py
+++ b/ooni/deck.py
@@ -205,7 +205,7 @@ class Deck(InputFile):
             if net_test_loader.requiresTor:
                 self.requiresTor = True
         except e.MissingTestHelper:
-            if self.preferred_backend is "onion":
+            if self.preferred_backend == "onion":
                 self.requiresTor = True
 
         self.netTestLoaders.append(net_test_loader)


bassosimone:
Oh my god! This is very anti-intuitive!

bassosimone:
I see a failed build in the `pr` but not in the `push`, hence I am going to assume that the failure was caused by a transient network error. Also, given the simplicity of the proposed changed it would be strange for it to be the culprit of the failure only for Twisted 14.0 and only in one of the two builds that were triggered by this pull request.

All the above is a very long way of saying *merging*!

tumblr_inline_ni5dnyog2g1rxr4x6

bassosimone:
Uhm, it seems I'm not super good at flaying Tardis! So, [this time another version of Twisted failed](https://travis-ci.org/TheTorProject/ooni-probe/jobs/145946826)... @hellais, can you confirm this is a transient issue or could this failure (apparently equal to the previous) symptom of a real issue?

hellais:
@bassosimone yeah it's not a "serious" failure. They are just failures caused by a dirty reactor. These happen when some `callLater` calls are left hanging without them triggering the the callback or expiring.
Since they are due to timing they sometimes appear and sometimes they don't.
I will nonetheless file an issue about these, since it's quite annoying that the unittests will fail sometimes due to them.
#572

PR #574, opened 2016-07-24
Add backend to required options to show error when it's not provided

hellais:
* Cast package summary to string


hellais:
I am merging this as it's an important fix.

PR #581, opened 2016-08-03
Feature/webui2 rb

hellais:
This is equivalent to the PR in https://github.com/TheTorProject/ooni-probe/pull/567, but some rebasing, cherry-picking trickery has been done to minimize the diffs.

In particular it's targeting the branch v2.0.0-alpha that includes most of the cleanup commits, while here is the juice of the webUI branch.


hellais:
I am going to say we do a feature freeze on this branch and proceed with only fixing bugs on it.

The idea is to leave it in staging on this branch until monday, then I am going to merge it into the alpha testing branch and tag the first alpha version of it and make a call for testing with a built package of it.

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -77,6 +77,12 @@ class WebConnectivityTest(httpt.HTTPTest, dnst.DNSTest):
     requiresTor = False
     followRedirects = True
 
+    # These are the options to be shown on the GUI
+    simpleOptions = [
+        {"name": "url", "type": "text"},
```
bassosimone: I see from below that the `file/url` type exists. So, why isn't a `url` field of type `file/url`?

hellais: Because this is actually not a file, but a single URL to be put inside of an `<input type='text'>`.

Vagrantfile
```diff
@@ -13,10 +13,16 @@ apt-get update
...
-apt-get install -y build-essential libdumbnet-dev libpcap-dev libgeoip-dev libffi-dev python-dev python-pip
+apt-get install -y build-essential libdumbnet-dev libpcap-dev libgeoip-dev libffi-dev python-dev python-pip libssl-dev
 cd /data/ooni-probe
 python setup.py install
 
+
+echo "To run ooniprobe:"
+echo "1. vagrant ssh probe"
+echo "2. ooniprobe-agent start"
+echo "3. connect to the web UI on your host machine at http://localhost:8042/"
```
bassosimone: Since this is running on a Vagrant, there should be no browser. Does `ooniprobe-agent` listen on default only on the localhost or on any interface? Probably it should listen on localhost by default and on any interface only if enabled by the user. In any case, I guess this instructions above require attention.

bassosimone: Ah, never mind, now I see below that the port is forwarded.

ooni/ui/cli.py
```diff
@@ -428,7 +463,7 @@ def runWithDaemonDirector(global_options):
     except ImportError:
         print "Pika is required for queue connection."
         print "Install with \"pip install pika\"."
-        sys.exit(7)
+        raise SystemExit(7)
```
bassosimone: What is the difference between the two?

ooni/ui/web/client/index.html
```diff
@@ -13,5 +13,5 @@
     <app>
       Loading...
     </app>
-  </body>
+  <script type="text/javascript" src="app.bundle.js?f06164bd3b339e781c75"></script></body>
```
bassosimone: What is the meaning of the query string?

ooni/ui/cli.py
hellais: Actually there is no difference, but I figured it made it a bit easier to scan the code looking for raises rather than looking for a `sys.exit`.

ooni/ui/web/client/index.html
hellais: Actually that is something generated automatically by webpack when it builds a new version of the app.bundle.js file. The reason why it appends to the app.bundle.js src file a random query string is to bust the cache of browsers that otherwise would not be serving you the newly built file.

hellais:
I am going to merge this into the 2.0.0 branch so I can tag an alpha release.

PR #582, opened 2016-08-03
V2.0.0 alpha

hellais:
This is the work in progress branch to keep track of the status of ooniprobe alpha.

We are probably not going to be integrating this into master sooner than end of August, early September. So the purpose of this is give people a tip that they can use to test ooniprobe 2.0.0 and keep track of when it's ready to be finally integrated into master.


matcheydj:
versions.py?

Vagrantfile
```diff
@@ -13,10 +13,16 @@ apt-get update
...
 
 # Setup for sniffer subsystem
-apt-get install -y build-essential libdumbnet-dev libpcap-dev libgeoip-dev libffi-dev python-dev python-pip
+apt-get install -y build-essential libdumbnet-dev libpcap-dev libgeoip-dev libffi-dev python-dev python-pip libssl-dev
 cd /data/ooni-probe
 python setup.py install
 
+
+echo "To run ooniprobe:"
+echo "1. vagrant ssh probe"
```
bassosimone: Bikeshedding: do we want to include `vagrant up probe` as step zero?

Vagrantfile
```diff
@@ -47,13 +53,21 @@ cp oonib.conf.example /etc/oonibackend.conf
 echo "Installing ooni-backend"
 python setup.py install
 
+echo ""
+echo "To run oonibackend:"
+echo "1. vagrant ssh backend"
```
bassosimone: ditto

Vagrantfile
```diff
@@ -67,21 +81,9 @@ Vagrant.configure("2") do |config|
...
-      echo "To run oonibackend:"
-      echo "1. vagrant ssh backend"
-      echo "2. vi /etc/oonibackend.conf  # possibly"
-      echo "3. cd /data/ooni-backend"
-      echo "4. sudo ./bin/oonib -c /etc/oonibackend.conf"
-    EOF
+  config.vm.define "testing" do |testing|
+    testing.vm.network "forwarded_port", guest: 8842, host: 8142
+    testing.vm.synced_folder ".", "/data/ooni-probe"
   end
```
bassosimone: what is the purpose of the testing box?

data/decks/web-full.yaml
```diff
@@ -0,0 +1,23 @@
+---
+name: Full Web test deck
+description: This deck runs HTTP Header Field Manipulation, HTTP Invalid
+    Request Line and the Web Connectivity test
+schedule: "@daily"
```
bassosimone: Is this quoted because a string cannot start with @ in yaml?

data/ooniprobe.conf.sample
```diff
@@ -5,6 +5,7 @@
 basic:
     # Where OONIProbe should be writing it's log file
```
bassosimone: should be `its`

data/updater.py
```diff
@@ -0,0 +1,357 @@
+#!/usr/bin/env python2
+
```
bassosimone: This is the update meant only for updating lepidopter, correct? If so, perhaps, I'd specify that with a toplevel comment (or docstring) in the file.

bassosimone: And/or, given the place where this script looks for configuration files, it would perhaps be better to just give it as name `lepidopter-update.py`

data/updater.py
```diff
@@ -0,0 +1,357 @@
...
+        except CalledProcessError:
+            raise InvalidPublicKey
+
+        try:
+            output = check_output(["gpg", "--batch", "--status-fd", "1",
+                                   "--no-default-keyring", "--keyring",
+                                   tmp_key, "--trust-model", "always",
+                                   "--verify", signature_path, file_path])
+        except CalledProcessError:
+            raise InvalidSignature
```
bassosimone: Notye to self: understand what these two commands do?

data/updater.py
```diff
@@ -0,0 +1,357 @@
...
+
+    return output
+
+class UpdateFailed(Exception):
+    pass
+
+def perform_update(version, skip_verification=False):
+    try:
+        updater = get_request(UPDATE_BASE_URL + "{0}/update.py".format(version))
+        updater_path = os.path.join(UPDATER_PATH, "update-{0}.py".format(version))
```
bassosimone: ah, so the scoping rules of python with respect to try...except blocks are that you can define a variable inside such a block and then use it outside!

data/updater.py
```diff
@@ -0,0 +1,357 @@
...
+        except OSError as ose:
+            if ose.errno != errno.EEXIST:
+                raise
+
+    with open(CURRENT_VERSION_PATH, "w") as out_file:
+        out_file.write("0")
+
+    # Copy myself over to the SCRIPT_INSTALL_PATH
+    shutil.copyfile(__file__, SCRIPT_INSTALL_PATH)
+    os.chmod(SCRIPT_INSTALL_PATH, int('744', 8))
```
bassosimone: Ah, from what I recall, I believe this `int('744', 8))` as an attempt to be compatible with python3. Has this script been tested with python3 as well?

data/updater.py
```diff
@@ -0,0 +1,357 @@
...
+
+    with open(PUBLIC_KEY_PATH, "w") as out_file:
+        out_file.write(PUBLIC_KEY)
+    os.chmod(PUBLIC_KEY_PATH, int('644', 8))
```
bassosimone: wouldn't it make sense to enforce root:root ownership?

data/updater.py
```diff
@@ -0,0 +1,357 @@
...
+
+    with open(SYSTEMD_SCRIPT_PATH, "w") as out_file:
+        out_file.write(SYSTEMD_SCRIPT)
+
+    check_call(["systemctl", "enable", "lepidopter-update"])
+    check_call(["systemctl", "start", "lepidopter-update"])
```
bassosimone: What does systemd if the service was already installed and running when you call these commands? I assume, given the code, that it would do the right thing, anyway better to ask.

Vagrantfile
hellais: I should probably delete this, I committed it by mistake. I was using the testing box as a machine I would snapshot and attempt installs on a clean system and on a system with a particular setup.

data/decks/web-full.yaml
hellais: Yeah `@` is a indicator character that according to the [spec](http://www.yaml.org/spec/1.2/spec.html#id2772075) is "reserved for future use".

data/ooniprobe.conf.sample
hellais: good call.

data/updater.py
hellais: The basically convert an ascii armored key into a binary key that can be used as a keyring so that we can verify the file by pointing to the binary key as a keyring.

data/updater.py
hellais: Yeah, it can however bite you quite badly if you are not raising the exception in the except clause as it can lead to an undefined variable if the except clause doesn't define those variables.

data/updater.py
hellais: That is mainly to make it easier to understand the permissions you set, since usually you reason in terms of base8 permissions (like when you run chmod from the command line), while os.chmod takes a integer. It's more readable to see `os.chmod(xx, int('744', 8))` than `os.chmod(xx, 484)`

data/updater.py
hellais: I guess this is kind of implicit from the fact that you are to run this from a root user. Maybe it's a good idea to check that when the script runs though.

data/updater.py
hellais: Yes, re-running enable and start operations on an already started or enabled service just does nothing and exits with exit status 0.

data/updater.py
hellais: Yes that makes sense. Will do.

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+
+from ooni.geoip import probe_ip
+from ooni.measurements import list_measurements
+
+class FileSystemlockAndMutex(object):
+    """
+    This is a lock that is both a mutex lock and also on filesystem.
+    When you acquire it, it will first block on the mutex lock and then
+    once that is released it will attempt to acquire the lock on the
+    filesystem.
```
bassosimone: I am not sure the sentence above is correct. AFAICT the code does the following:

```
lock_mutex()
lock_fs()
do_something()
unlock_fs()
unlock_mutex()
```

but the sentence says "once [the mutex] is released it will attempt to acquire the lock on the fileystem. I think it should say instead, "once [the mutex] is acquired". Correct?

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+    _time_format = "%Y-%m-%dT%H:%M:%SZ"
+    schedule = None
+    identifier = None
+
+    def __init__(self, schedule=None, identifier=None,
+                 scheduler_directory=config.scheduler_directory):
+        if schedule is not None:
+            self.schedule = schedule
+        if identifier is not None:
+            self.identifier = identifier
```
bassosimone: What is the difference between what you do above and just declaring `self.schedule = schedule` and `self.identifier = identifier` here in `__init__()`?

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+    def cancel(self):
+        self._last_run_lock.release()
+
+    @property
+    def should_run(self):
+        current_time = datetime.utcnow().replace(tzinfo=tz.tzutc())
+        next_cycle = croniter(self.schedule, self.last_run).get_next(datetime)
+        if next_cycle <= current_time:
+            return True
+        return False
```
bassosimone: Couldn't this become `return next_cycle <= current_time`?

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+    well.
+    """
+    identifier = 'send-heartbeat'
+    schedule = '@hourly'
+
+    def task(self):
+        # XXX implement this
+        pass
+
+# Order mattters
```
bassosimone: One excess `t` here

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+]
+
+
+@defer.inlineCallbacks
+def run_system_tasks(no_input_store=False):
+    task_classes = SYSTEM_TASKS[:]
+
+    if no_input_store:
+        log.debug("Not updating the inputs")
+        task_classes.remove(UpdateInputsAndResources)
```
bassosimone: For robustness with respect to future changes, I'd wrap this `remove` with `try...except`

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+
+    for task_class in task_classes:
+        task = task_class()
```
bassosimone: This means that we can only create tasks that do not receive explicit arguments, right?

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+        self.interval = interval
+        self._looping_call = LoopingCall(self._should_run)
+        self._scheduled_tasks = []
+
+    def schedule(self, task):
+        self._scheduled_tasks.append(task)
+
+    def unschedule(self, task):
+        # We first cancel the task so the run lock is deleted
+        task.cancel()
```
bassosimone: Note to self: verify what happens if we `cancel()` a task that is currently running

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+        self._scheduled_tasks.remove(task)
+
+    def refresh_deck_list(self):
```
bassosimone: It seems to me that this function runs decks, so perhaps I don't fully understand its name. Does it mean that the function objective is to refresh the [enabled] decks list?

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,409 @@
...
+        log.debug("Ran {0}".format(task.identifier))
+
+    def _should_run(self):
+        """
+        This function is called every self.interval seconds to check
+        which periodic tasks should be run.
+        """
+        for task in self._scheduled_tasks:
+            log.debug("Running task {0}".format(task.identifier))
+            d = task.run()
```
bassosimone: What happens if the configuration is crazy and tasks are re-run when they are running? The lock mechanism should protect against that, isn't it?

ooni/deck/backend.py
```diff
@@ -0,0 +1,191 @@
...
+        raise e.NoReachableTestHelpers
+    else:
+        defer.returnValue(test_helper_address.encode('ascii'))
+
+
+@defer.inlineCallbacks
+def get_reachable_collector(collector_address, collector_alternate,
+                            preferred_backend):
+    # We prefer onion collector to https collector to cloudfront
+    # collectors to plaintext collectors
```
bassosimone: I think this comment should be moved above at line 19 of this file when the vector containing the prioritized address types is defined.

ooni/deck/backend.py
```diff
@@ -0,0 +1,191 @@
...
+    def find_collector_and_test_helpers(test_name, test_version):
+        # input_files = [u""+x['hash'] for x in input_files]
+        for net_test in provided_net_tests:
+            if net_test['name'] != test_name:
+                continue
+            if net_test['version'] != test_version:
+                continue
+            # XXX remove the notion of policies based on input file hashes
+            # if set(net_test['input-hashes']) != set(input_files):
+            #    continue
```
bassosimone: So, this basically means that input_hashes are no longer honoured?

ooni/deck/deck.py
```diff
@@ -0,0 +1,411 @@
...
+from ooni.settings import config
+from ooni.utils import log, generate_filename
+
+
+def resolve_file_path(v, prepath=None):
+    from ooni.deck.store import input_store
+    if v.startswith("$"):
+        # This raises InputNotFound and we let it carry onto the caller
+        return input_store.get(v[1:])["filepath"]
+    elif prepath is not None and (not os.path.isabs(v)):
```
bassosimone: Since above you do return here it suffices to use `if`

ooni/director.py
```diff
@@ -84,13 +91,69 @@ def __init__(self):
...
+
+    def subscribe(self, handler):
+        self._subscribers.append(handler)
+
+    def unsubscribe(self, handler):
+        self._subscribers.remove(handler)
+
+    def notify(self, event):
+        for handler in self._subscribers:
+            handler(event)
```
bassosimone: Does it make sense here to `try...catch` around calling `handler` such that all the subscribers are iterated even if some of them fail?

ooni/nodes.conf
```diff
@@ -1,53 +0,0 @@
-[control]
```
bassosimone: What was this file?

ooni/resources.py
```diff
@@ -0,0 +1,184 @@
...
+import errno
+
+from twisted.python.filepath import FilePath
+from twisted.internet import defer
+from twisted.web.client import downloadPage, getPage, HTTPClientFactory
+
+# WARNING: this script is being run as part of the post install procedure.
+# Be sure to not import either in this module or in the imported modules
+# dependencies other than twisted. If you end up including something that is
+# not twisted, then you will need to add it to the setup_requires in setup.py.
```
bassosimone: Shouldn't this be in the script folder?

ooni/ui/cli.py
```diff
@@ -2,7 +2,6 @@
```
bassosimone: I do not fully understand the difference in terms of taxonomy between `ooni/foo/cli.py` and `ooni/script/foo.py`. Is the former the implementation and the latter just the "main"?

ooni/ui/web/client/index.html
```diff
@@ -0,0 +1,17 @@
...
+    <title></title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+  </head>
+  <body ng-app="app" ng-strict-di ng-cloak>
+
+    <app>
+      Loading...
```
bassosimone: So, if I recall correctly, when one sees this page it means that the web-ui was not correctly installed... would it make sense, then, to state that here and add instructions how to get the ui?

bassosimone:
# Review tasks
- [ ] first reading of the full diff
- [ ] second, more in depth reading
- [ ] light testing (others have been testing already)
- [ ] merging

ooni/agent/scheduler.py
hellais: We do this check because the class attributes are override inside of the subclasses of ScheduledTask and we don't want the constructor to rewrite the instance attribute with `None` when a not `None` value is passed.

ooni/agent/scheduler.py
hellais: Yes that would also be equivalent, though I think it's a bit more clear if it's explicit.

ooni/agent/scheduler.py
hellais: It matters a lot :P

ooni/agent/scheduler.py
hellais: Yeah, that makes sense.

ooni/agent/scheduler.py
hellais: For system tasks, yes.

ooni/agent/scheduler.py
hellais: It actually doesn't run them, but it just ensures that the ones that are set as enabled are configured to run periodically.

ooni/agent/scheduler.py
hellais: Exactly, the locking mechanism is in place to ensure that it will lock when somebody has already triggered a run.

There is a tradeoff that should probably be noted somewhere, that this backlog of tasks waiting on the lock can be quite large for a task that requires a long time to complete running.

For example if a task requires 1 hour to complete you will have a backlog of 60*60/30= 120 run operations.

ooni/deck/backend.py
hellais: Yes, that is correct. We don't actually have any active collector that is using them anyways and it's a feature that I think we should entirely scrap out from collectors in future versions, since it really doesn't protect against a sufficiently motivated adversary (they can just pretend an input file has a certain hash) and it leads to reading the whole and computing a hash on it before running the test.

ooni/director.py
hellais: Yes I think this is a safer behavior.

ooni/nodes.conf
hellais: This is some leftover from super old legacy code that had the notion of remotely administrable ooniprobe nodes that would be used for running tests on.

No code using it exists anymore and we don't have this type of usecase in mind anymore.

ooni/resources.py
hellais: This comment should actually be deleted since it's no longer accurate. The setup.py doesn't import the resources anymore and doesn't run it as part of the postinstall.

ooni/ui/cli.py
hellais: Yes correct. Everything inside of ooni/script/* is also an actual script that you can run indipendently and it will behave as a standard python script with an entry point called `run()`.

What is inside of `ooni/foo/cli.py` are the functions related to supporting the execution of the script.

ooni/ui/web/client/index.html
hellais: Well that is actually displayed for some seconds while the JS resources are fetched and loaded. I agree though that we could perhaps register some exception handlers on the loading of the assets and if it gets triggered also show some instructions on how to set it up.

I plan however on adding the source for the web app to the repo every time a stable release is tagged so that people can also use it from git builds.

hellais:
Thanks for the very in depth review. I will begin addressing some of your comments in upcoming commits.

data/updater.py
bassosimone: ah, I see!

data/updater.py
bassosimone: Yes, I see. My point was that `0644` is valid python2 while in python3 you would need to write `0o644` instead (don't ask me why did they do this when `0644` is pretty much what one would expect). So, the portable way of writing `0644` is indeed to do as you did, i.e. convert from the string.

ooni/agent/scheduler.py
bassosimone: Right. Given that those are classes attributes and not instance attributes, this means that using them differently for each instance would only work as long as non mutables are assigned to the class attributes, correct?

ooni/agent/scheduler.py
bassosimone: Meaning that, in other words, there is the implicit assumption that system tasks do not need any explicit parameter passed to their constructor?

ooni/resources.py
bassosimone: Who runs this script then? It seems quite detached from the rest of ooni, isn'it?

ooni/ui/web/client/index.html
bassosimone: :+1: for adding the web app to allow people to use it from git builds, I think this makes easier for people to test ooni-probe. With this change in place the usefulnees of an event handler is reduced, because, if it does not work, it's because it's broken and not because of user-that-did-not-downloaded-the-latest-gui error.

Besides, how do I fetch the latest version of the gui for testing purposes?

ooni/resources.py
hellais: The use of the term "script" is actually very misleading here. It's actually a module used (in particular the `check_for_updates` function) by the scheduled tasks responsible for updating the inputs and resources.

ooni/agent/scheduler.py
hellais: I am not sure I understand what you say here. The reason for supporting also passing in arguments is that, the subclass should be able to specify what is the default schedule, but a user of the task may decide to override it with something else at runtime. Same thing goes for the identifier, that in some cases needs to be specialised (for example with the RunDeck task).

ooni/ui/web/client/index.html
hellais: Currently we don't have a tagged version of the latest gui. For testing you should build it yourself from the master.

How I do this while developing the web-ui is that I build it using:

```
gulp build --watch --dest /path/to/ooni-probe/ooni/ui/web/client/
```

bassosimone: Thank you!

ooni/agent/scheduler.py
bassosimone: Probably it's not fully clear to me how class attributes are not shared by all instances of a class but rather become in some way specific of a specific instance.

hellais: Let me see if I can give you a couple of examples as to why this is like this and perhaps you have some idea on how the code can be improved.

There are mainly 2 types of uses for the constructor of this class:

1. We use the default class attributes for schedule and identifier (this is the most common case), for example with the following:

* `UploadReports` (https://github.com/TheTorProject/ooni-probe/blob/v2.0.0-alpha/ooni/agent/scheduler.py#L414 and https://github.com/TheTorProject/ooni-probe/blob/v2.0.0-alpha/ooni/agent/scheduler.py#L149)
* `RefreshDeckList` (https://github.com/TheTorProject/ooni-probe/blob/v2.0.0-alpha/ooni/agent/scheduler.py#L263 and https://github.com/TheTorProject/ooni-probe/blob/v2.0.0-alpha/ooni/agent/scheduler.py#L417)

2. We need to change the identifier and the schedule depending on some value we read.
This is actually only the case for the RunDecks where the schedule and identifier is determined based on the deck name and the schedule written inside of the deck itself: https://github.com/TheTorProject/ooni-probe/blob/v2.0.0-alpha/ooni/agent/scheduler.py#L384
For this reason we need to support specialising at construction time these class attributes.

bassosimone: Okay, let me see if now I understand what is happening here: in case 1 those two attributes are proper class attributes whereas in case 2 they are instance attributes (meaning that in case 2 each instance gets different values for those attributes), is this correct?

With this in mind, I now understand how the code is written as it is. Now, I have a question: how does class attributes work? When the class is created, they are copied in the `__dict__` underlying a Python instance? Or does `__getattr__` look into the instance `__dict__` and, if that fails, it then looks into the class `__dict__`? I would say the former, because otherwise I do not clearly see how different instances could have different values for these attributes.

To say the same with other words: I find it strange that you can modify "class" attributes for specific instances without affecting other instances. How is that possible? I was thinking that class attributes are like the `prototype` in JavaScript in that the same values of the class attributes are shared by all instances of a class, but apparently it's not like so.

/me still puzzled

ooni/agent/scheduler.py
```diff
@@ -0,0 +1,423 @@
...
+        self._fs_lock.unlock()
+        self._mutex.release()
+
+    @property
+    def locked(self):
+        return self._mutex.locked or self._fs_lock.locked
+
+# We use this date to indicate that the scheduled task has never run.
+# Easter egg, try to see what is special about this date :)?
+CANARY_DATE = datetime(1957, 8, 4, tzinfo=tz.tzutc())
```
bassosimone: Have not found the easter egg yet, but this date is before the Epoch. Are we sure there are not going to be any implications in this?

ooni/agent/scheduler.py
bassosimone: @hellais what happens here if the task is running?

ooni/agent/scheduler.py
bassosimone: > There is a tradeoff that should probably be noted somewhere [...]

True. I think adding a comment would be just fine.

ooni/backend_client.py
```diff
@@ -212,105 +214,12 @@ def err(failure):
...
-                input_file.verify()
-                return input_file
-
-            @d.addErrback
-            def err(err):
-                log.err("Failed to download the input file %s" % input_hash)
-                log.exception(err)
-
-            return d
-
```
bassosimone: @hellais was the code above move somewhere else or just removed?

ooni/backend_client.py
```diff
@@ -212,105 +214,12 @@ def err(failure):
...
-                deck.verify()
-                return deck
-
-            @d.addErrback
-            def err(err):
-                log.err("Failed to download the deck %s" % deck_hash)
-                log.exception(err)
-
-            return d
-
```
bassosimone: @hellais was the code above removed or moved?

ooni/agent/scheduler.py
hellais: Previously we would have had epoch 0 in here, but I thought it would be better to have a date that is actually specific to ooniprobe to distinguish it from a time that may be set when a system boots up with 0 seconds from epoch and hence leading to a timestamp of 0 (that would mean we would trigger the first_run) twice.

Since we manage the time always ourselves and serialize it using methods that support times before epoch we should be all good.

ooni/measurements.py
```diff
@@ -0,0 +1,145 @@
...
+    keep = False
+    stale = False
+    if measurement.child("measurements.njson.progress").exists():
+        completed = False
+        # XXX this is done quite often around the code, probably should
+        # be moved into some utility function.
+        pid = measurement.child("running.pid").open("r").read()
+        pid = int(pid)
+        try:
+            os.kill(pid, signal.SIG_DFL)
```
bassosimone: What is the purpose of this? Identifying a measurement run by another instance of ooni-probe? (afaict ooni-probe does not fork to run a new measurement)

ooni/measurements.py
```diff
@@ -0,0 +1,145 @@
...
+            running = True
+        except OSError:
```
bassosimone: What happens if you end up killing a pid owned by another user? Asking because I am wondering whether it would be more robust to check the error code here.

ooni/measurements.py
hellais: Because if you run the ooniprobe cli you want to see if the process that was running that measurement is still running and therefore it will eventually complete it.

ooni/measurements.py
```diff
@@ -0,0 +1,145 @@
...
+        except:
+            log.err("Failed to get metadata for measurement {0}".format(measurement_id))
+    return measurements
+
+if __name__ == "__main__":
+    import sys
+    if len(sys.argv) != 3:
+        print("Usage: {0} [input_file] [output_file]".format(sys.argv[0]))
+        sys.exit(1)
+    generate_summary(sys.argv[1], sys.argv[2])
```
bassosimone: I like adding minimal main()s to modules to make them callable for testing

ooni/nettest.py
```diff
@@ -189,21 +194,22 @@ def __init__(self, options, test_file=None, test_string=None,
...
-            'probe_cc': config.probe_ip.geodata['countrycode'],
-            'probe_ip': config.probe_ip.geodata['ip'],
-            'probe_city': config.probe_ip.geodata['city'],
+            'probe_asn': probe_ip.geodata['asn'],
+            'probe_cc': probe_ip.geodata['countrycode'],
+            'probe_ip': probe_ip.geodata['ip'],
+            'probe_city': probe_ip.geodata['city'],
             'software_name': 'ooniprobe',
             'software_version': ooniprobe_version,
+            # XXX only sanitize the input files
```
bassosimone: I do not fully understand this comment

ooni/nettest.py
```diff
@@ -231,28 +237,14 @@ def _accumulateInputFiles(self, test_class):
...
-
             'filename': None
         }
         m = ONION_INPUT_REGEXP.match(filename)
         if m:
-            input_file['url'] = filename
-            input_file['address'] = m.group(1)
-            input_file['hash'] = m.group(2)
+            raise e.InvalidInputFile("Input files hosted on hidden services "
+                                     "are not longer supported")
```
bassosimone: s/not/no/g

ooni/measurements.py
hellais: You get an `OSError` with `errno` set to `errno.EPERM` when you don't have permissions, while when it doesn't exist you get `OSError` with `errno.ESRCH`.

Maybe I can make this a bit tighter.

ooni/backend_client.py
hellais: This is part of a feature set that I don't plan on supporting anymore in the future. What is allowed to do was download inputs and decks based on their hash from the collector/bouncer.

Since we now have support for downloading inputs via the resource update mechanism this should be dropped.

ooni/measurements.py
hellais: I think this can actually be removed as I put it there while doing direct testing of this functionality by invoking the script standalone from the command line, but we shouldn't actually support this usage.

ooni/agent/scheduler.py
bassosimone: Right, thanks

ooni/measurements.py
bassosimone: right, makes sense

ooni/nettest.py
hellais: Yeah this is a note for something that needs to be done, that is not currently done. Basically we treat all options as if they were filepaths and sanitize them in the sense that we strip the prefix to avoid leaking filesystem paths of the user. This can have unintended consequences if the option looks like a filepath, but actually isn't leading to it being stripped.

ooni/measurements.py
bassosimone: Yes, because, in light of your explanation, I wonder what happens if you're $your_user and attempt to kill the daemon running as $ooni_user

ooni/nettest.py
```diff
@@ -454,7 +446,11 @@ def checkAllTasksDone(self):
                   (self.doneTasks, self.tasks))
         if self.completedScheduling and \
                 self.doneTasks == self.tasks:
-            self.allTasksDone.callback(self.doneTasks)
+            if self.allTasksDone.called:
+                log.err("allTasksDone was already called. This is probably a bug.")
```
hellais: Yeah a stacktrace here doesn't really provide much insight into the problem and it was the case in the past that we would throw it (in the form of an already called error).

I believe this problem should be resolved, but I am leaving it here to see if actually it's not fixed and so people can report this if it pops up in their error log.

ooni/measurements.py
hellais: Yes, it should also consider it to be running if it fails due to errno.EPERM. Pushed a fix for this and abstracted it into a function.

ooni/agent/scheduler.py
hellais: If the task is running it will actually continue running until the end. There is currently no concept of actually stopping a running task and things get even more tricky when a deck is concerned.

There is a ticket open regarding this, but it's deferred to future versions: https://github.com/TheTorProject/ooni-probe/issues/590

bassosimone: Right. My only concern left is whether releasing the combined lock is idempotent; i.e. if releasing a non-locked or already-released lock does nothing.

ooni/reporter.py
```diff
@@ -206,6 +208,56 @@ def createReport(self):
...
+            'input': e.pop('input', None),
+            'id': str(uuid.uuid4()),
+            'test_start_time': e.pop('test_start_time', None),
+            'measurement_start_time': e.pop('measurement_start_time', None),
+            'test_runtime': e.pop('test_runtime', None),
+            'test_keys': e
+        }
+        report_entry.update(self.testDetails)
+        self._write(json.dumps(report_entry))
+        self._write("\n")
```
bassosimone: Here you can probably use `_writeln()` which otherwise seems unused

ooni/reporter.py
```diff
@@ -313,162 +363,176 @@ def finish(self):
...
+        incomplete_reports = []
+        all_entries = yield self.get_report_log_entries()
+        for entry in all_entries[:]:
+            # This means that the measurement itself is incomplete
+            if entry['completed'] is False:
+                continue
+            if entry['status'] in ('created',):
+                try:
+                    os.kill(entry['pid'], 0)
+                except OSError:
```
bassosimone: Maybe also here you would like to distinguish between different reasons why `kill()` failed?

ooni/reporter.py
```diff
@@ -313,162 +363,176 @@ def finish(self):
...
+        in_progress_reports = []
+        all_entries = yield self.get_report_log_entries()
+        for entry in all_entries[:]:
+            if entry['status'] in ('created',):
+                try:
+                    os.kill(entry['pid'], 0)
+                    in_progress_reports.append(
+                        (entry['measurements_path'], entry)
+                    )
+                except OSError:
```
bassosimone: And also here?

ooni/reporter.py
```diff
@@ -313,162 +363,176 @@ def finish(self):
...
+                    pass
+        defer.returnValue(in_progress_reports)
+
+    @defer.inlineCallbacks
+    def get_to_upload(self):
+        to_upload_reports = []
+        all_entries = yield self.get_report_log_entries()
+        for entry in all_entries[:]:
+            # This means that the measurement itself is incomplete
```
bassosimone: I do not understand this comment

ooni/resources.py
```diff
@@ -0,0 +1,179 @@
...
+            try:
+                dst_file.parent().makedirs()
+            except OSError as e:
+                if not e.errno == errno.EEXIST:
+                    raise
+            src_file = dst_file.temporarySibling()
+            src_file.alwaysCreate = 0
+
+            temporary_files.append((dst_file, src_file))
+            # The paths for the download require replacing "/" with "."
```
bassosimone: I think this part of the protocol is quite fragile. What about specifying the desired directory relative to the resources folder separately from the file name.

ooni/scripts/oonideckgen.py
```diff
@@ -0,0 +1,138 @@
...
+         "generate the deck."],
+        ["collector", None, None, "Specify a custom collector to use when "
+                                  "submitting reports"],
+        ["bouncer", None, None, "Specify a custom bouncer to use"],
+        ["output", "o", None,
+         "Specify the path where we should be writing the deck to."]
+    ]
+
+    def opt_version(self):
+        print("oonideckgen version: " % __version__)
```
bassosimone: I find it confusing that different commands part of ooni have different version numbers. I'd be less confused if either of this happens:

1. `oonideckgen version: 0.2.0-rc.2`

2. `oonideckgen version: 1.0.0 (ooni/0.2.0-rc2.)`

ooni/reporter.py
hellais: Yes I will use the is_process_running function here too.

ooni/scripts/oonideckgen.py
```diff
@@ -0,0 +1,138 @@
...
+
+    print("Deck written to {0}".format(options['output']))
+    print("Run ooniprobe like so:")
+    print("ooniprobe -i {0}".format(options['output']))
+
+
+@defer.inlineCallbacks
+def get_user_country_code():
+    config.privacy.includecountry = True
+    yield probe_ip.lookup()
```
bassosimone: Uhm, would it be possible to pass to `probe_ip.lookup()` a flag to force it to return the country rather than changing a global value?

I am wondering this could break things if, in some way in the future, `config.privacy` is somehow written to disk and hence setting it in this context is going to become persistent.

ooni/scripts/oonideckgen.py
```diff
@@ -0,0 +1,138 @@
...
+        print("%s: --country-code must be 2 characters" % sys.argv[0])
+        sys.exit(2)
+
+    if os.path.isdir(options['output']):
+        options['output'] = os.path.join(options['output'], 'web-full.yaml')
+
+    options['country-code'] = options['country-code'].lower()
+
+    try:
+        os.makedirs(os.path.dirname(options['output']))
```
bassosimone: You may want an utility function that does that. I've seen this pattern used more than once throughout the code now.

ooni/scripts/oonideckgen.py
```diff
@@ -0,0 +1,138 @@
...
+    except OSError as exception:
+        if exception.errno != errno.EEXIST:
+            raise
```
bassosimone: Or perhaps *this* pattern, including the error check

ooni/reporter.py
hellais: On second thought my concern with using `_writeln` is that it's doing string concatenation that can be expensive if the string to be concatenated is fairly large.

I will change how `writeln` works to be better at this.

ooni/scripts/oonideckgen.py
```diff
@@ -0,0 +1,138 @@
...
+
+    generate_deck(options)
+
+def run():
+    task.react(oonideckgen)
```
bassosimone: Wow, how would Twisted know that it's time to exit the reactor when it reaches the end of the task? That is, how is the end of the task defined?

ooni/reporter.py
hellais: What this means ifs that if it's not complete it means that it's incomplete (I guess by definition) and hence we shouldn't be uploading it as it's a partial measurement.

ooni/scripts/ooniprobe.py
```diff
@@ -0,0 +1,35 @@
...
+
+def ooniprobe(reactor):
+    from ooni.ui.cli import runWithDaemonDirector, runWithDirector
+    from ooni.ui.cli import setupGlobalOptions, initializeOoniprobe
+
+    global_options = setupGlobalOptions(logging=True, start_tor=True,
+                                        check_incoherences=True)
+    if global_options['queue']:
+        return runWithDaemonDirector(global_options)
+    elif global_options['initialize']:
```
bassosimone: In this chain of `if..elif` you can omit the `elif` and go for `if` because you `return`

ooni/scripts/ooniprobe_agent.py
```diff
@@ -0,0 +1,193 @@
...
+                   '--umask', '022']
+
+    twistd_config = OoniprobeTwistdConfig()
+    if options is not None:
+        twistd_args.extend(options.twistd_args)
+    twistd_args.append("StartOoniprobeAgent")
+    try:
+        twistd_config.parseOptions(twistd_args)
+    except usage.error, ue:
+        print("ooniprobe: usage error from twistd: {}\n".format(ue))
```
bassosimone: Here you probably want to `raise`, right?

ooni/scripts/ooniprobe_agent.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    running_pidfile = None
+    for pidfile in [config.system_pid_path, config.user_pid_path]:
+        if not os.path.exists(pidfile):
+            # Didn't find the pid_file
+            continue
+        pid = open(pidfile, "r").read()
+        pid = int(pid)
+        if is_process_running(pid):
+            running_pidfile = pidfile
```
bassosimone: Why `else: continue` and not `break` just in this branch?

bassosimone: Or directly `return running_pidfile` and unconditional raise after the `for`?

ooni/scripts/ooniprobe_agent.py
```diff
@@ -0,0 +1,193 @@
...
+
+    pid = open(pidfile, "r").read()
+    pid = int(pid)
+    try:
+        os.kill(pid, signal.SIGKILL)
+    except OSError as ose:
+        if ose.errno == errno.ESRCH:
+            print("No process was running. Cleaning up.")
+            # the process didn't exist, so wipe the pid file
+            os.remove(pidfile)
```
bassosimone: It would probably be more robust to also wrap this `os.remove` with `try..except`

ooni/scripts/oonideckgen.py
hellais: It will exit the reactor loop and trigger a sys exit when the deferred returned by the function passed to react fires.

ooni/scripts/ooniprobe_agent.py
```diff
@@ -0,0 +1,193 @@
...
+    try:
+        get_running_pidfile()
+        print("ooniprobe-agent is running")
+        return 0
+    except NotRunning:
+        print("ooniprobe-agent is NOT running")
+        return 1
+
+def stop_agent():
+    # This function is borrowed from tahoe
```
bassosimone: I think a better way of implementing this function would be to send `SIGTERM` first, give the process some grace time to complete, then `SIGKILL`.

ooni/scripts/ooniprobe_agent.py
hellais: Yes fair point. Now that it has been refactored with a function to check if the process is running this can be rewritten as you suggest that is indeed much cleaner.

ooni/scripts/ooniprobe_agent.py
hellais: Ok I will change the behaviour to what you suggest.

ooni/scripts/oonideckgen.py
hellais: Yes I agree, maybe I will just change all the subcommands to follow the same versioning of ooniprobe itself.

ooni/scripts/ooniresources.py
```diff
@@ -38,7 +30,5 @@ def run():
         print "%s: Try --help for usage details." % (sys.argv[0])
         sys.exit(1)
 
-    if options['update-inputs'] or options['update-geoip']:
-        print("WARNING: Passing command line arguments is deprecated")
-
-    yield update.download_resources()
```
bassosimone: Uhm, I think that, if the script is *only* deprecated, it should keep downloading resources. Otherwise, you can make it a no-op for backwards compatibility, but in that case the message should be slightly different.

hellais: I actually don't want it to do anything since the management of the resource now is entirely delegated to the the scheduled tasks.

I am leaving the command though, since it may be part of peoples scripts and I don't want to break them. I am going to add to the output mention of the fact that it actually doesn't do anything.

ooni/ui/web/server.py
```diff
@@ -0,0 +1,608 @@
...
+
+    def get(self):
+        d = defer.Deferred()
+        self.deferred_subscribers.append(d)
+        return d
+
+class WebUIAPI(object):
+    app = Klein()
+    # Maximum number in seconds after which to return a result even if not
+    # change happenned.
```
bassosimone: "not" -> "no"; "happenned" = "happened"

ooni/ui/web/server.py
```diff
@@ -0,0 +1,608 @@
...
+    def __init__(self, config, director, scheduler, _reactor=reactor):
+        self._reactor = reactor
+        self.director = director
+        self.scheduler = scheduler
+
+        self.config = config
+        self.measurement_path = FilePath(config.measurements_directory)
+
+        # We use a double submit token to protect against XSRF
+        rng = SystemRandom()
```
bassosimone: Just a check: this is a secure RNG, right?

hellais: Yes, this uses underneath `/dev/urandom`

ooni/reporter.py
bassosimone: Okay, a better question would then be: what is the difference between `entry['completed']` and `entry['status']`?

ooni/scripts/oonideckgen.py
bassosimone: cool

ooni/utils/onion.py
```diff
@@ -212,6 +217,80 @@ def get_client_transport(transport):
...
+                else:
+                    bridges.append(bridge.strip())
+        tor_config.Bridge = bridges
+
+    if config.tor.torrc:
+        for i in config.tor.torrc.keys():
+            setattr(tor_config, i, config.tor.torrc[i])
+
+    if os.geteuid() == 0:
+        tor_config.User = pwd.getpwuid(os.geteuid()).pw_name
```
bassosimone: Does this mean that, if we're running as root, we run tor as root?

setup.py
```diff
@@ -69,58 +69,93 @@
...
+def install_lepidopter_update():
+    check_call(["data/lepidopter-update.py", "install"])
+
+
+def mkdir_p(path):
+    try:
+        os.makedirs(path)
+    except OSError as ose:
+        if ose.errno != errno.EEXIST:
+            raise
```
bassosimone: This duplicates stuff defined in ooni/utils.py... is this fine?

ooni/deck/deck.py
```diff
@@ -0,0 +1,411 @@
...
+            )
+            generate_summary(
+                measurement_dir.child("measurements.njson").path,
+                measurement_dir.child("summary.json").path
+            )
+            measurement_dir.child("running.pid").remove()
+
+    def _measurement_failed(self, failure, task):
+        if not task.output_path:
+            # XXX do we also want to delete measurements.njson.progress?
```
bassosimone: I think it may be useful to keep it around. Perhaps, a task that deletes `.progress` files after,  say, one week?

ooni/deck/legacy.py
```diff
@@ -0,0 +1,65 @@
...
+    new_deck_data["tasks"] = []
+    for deck_item in deck_data:
+        deck_task = {"ooni": {}}
+
+        options = deck_item["options"]
+        deck_task["ooni"]["test_name"] = options.pop("test_file")
+        deck_task["ooni"]["annotations"] = options.pop("annotations", {})
+        deck_task["ooni"]["collector"] = options.pop("collector", None)
+
+        # XXX here we end up picking only the last not none bouncer_address
```
bassosimone: @hellais can you shed light on what this comment means? In the old format was the `bouncer_address` value a list?

ooni/deck/store.py
```diff
@@ -0,0 +1,193 @@
...
+                    "id": "citizenlab_{0}_urls".format(cc),
+                    "type": "file/url"
+                }, out_fh)
+        self._cache_stale = True
+
+    @defer.inlineCallbacks
+    def create(self, country_code=None):
+        # XXX This is a hax to avoid race conditions in testing because this
+        #  object is a singleton and config can have a custom home directory
+        #  passed at runtime.
```
bassosimone: > Explaining a hack is like dissecting a frog, you understand it better but the frog dies in the process.

Nonetheless, could you explain the hack in here?

bassosimone: Ah, is the hack that you overwrite attributes set in the constructor? If so, good for the frog, that can safely leap away.

ooni/deck/store.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _update_cache(self):
+        for deck_path in self.available_directory.listdir():
+            if not deck_path.endswith('.yaml'):
+                continue
+            deck_id = deck_path[:-1*len('.yaml')]
+            deck = NGDeck(
+                deck_path=self.available_directory.child(deck_path).path
+            )
+            self._cache[deck_id] = deck
```
bassosimone: Uhm, shouldn't this function set `_cache_stale` to `False`?

Also, I do not understand why the function updating the cache does not restart from an empty cache (i.e. from `self._cache = {}`)

ooni/deck/store.py
```diff
@@ -0,0 +1,193 @@
...
+        mkdir_p(self.path.child("data").path)
+
+        yield self.update_url_lists(country_code)
+
+    @defer.inlineCallbacks
+    def update(self, country_code=None):
+        # XXX why do we make a difference between create and update?
+        yield self.create(country_code)
+
+    def _update_cache(self):
```
bassosimone: I am probably missing something regarding the cache. Here and also below you update the cache by inserting more items in the cache walking through a list, but I would have expected the cache to be cleared before updating it, such that items that are no longer on disk do not compare in cache.

ooni/deckgen/processors/namebench_dns_servers.py
```diff
@@ -1,51 +0,0 @@
-import os
```
bassosimone: This feature has been removed, right?

ooni/deckgen/processors/citizenlab_test_lists.py
```diff
@@ -1,55 +0,0 @@
-import os
```
bassosimone: this has been refactored into the deck itself, right?

ooni/deckgen/processors/namebench_dns_servers.py
hellais: Yes this is now deleted.

ooni/deckgen/processors/citizenlab_test_lists.py
hellais: Yes, it's handled in a very different way now, via the deck.

ooni/deck/store.py
hellais: Golly gosh, you are right. The cache must start from a clean state. Will update this to reflect that and ensure that the update is atomic.

ooni/deck/store.py
hellais: Yes good call.

ooni/deck/store.py
hellais: Yeah it's a hack because those attributes are also set in the constructor, but we we are unit-testing we override those global variables, but we override them after all the import chains have been run and the global deck_store is also created and therefore we need to reset them again here.

ooni/deck/legacy.py
hellais: Yeah, in the old format you could specify various bouncer addresses per individual test, which actually didn't make much sense since the lookup is only done to one bouncer.

ooni/deck/deck.py
hellais: Yes, that is something that happens. All measurements older than 7 days that are not explicitly kept will be deleted even if they have a .progress.

setup.py
hellais: It's better to not use inside of the setup.py stuff from your package as the package may have depedencies that are not met at install time.

ooni/utils/onion.py
hellais: Yes, that is correct. In the future we may try to be a bit smarter about this, but for the time being it's more robust to always ensure we run it as the same user as the current process.

ooni/ui/web/server.py
```diff
@@ -0,0 +1,608 @@
...
+            )
+
+        except usage.UsageError:
+            raise WebUIError(
+                400, 'Error in parsing options'
+            )
+
+        except errors.InsufficientPrivileges:
+            raise WebUIError(
+                400, 'Insufficient priviledges'
```
bassosimone: shouldn't it be `privileges` (i.e. without the d)?

ooni/ui/web/server.py
```diff
@@ -0,0 +1,608 @@
...
+        request.setHeader('Content-Type', 'text/plain')
+        request.setHeader('Content-Length', len(content))
+        return content
+
+    @app.route('/api/input/<string:input_id>', methods=["GET"])
+    @xsrf_protect(check=False)
+    @requires_true(attrs=['_is_initialized'])
+    def api_input_details(self, request, input_id):
+        return self.render_json(
+            self.director.input_store.get(input_id), request
```
bassosimone: Here I am not sure whether returning this input also returns the filepath, which above you try to hide

ooni/ui/web/server.py
```diff
@@ -0,0 +1,608 @@
...
+    def api_measurement_view(self, request, measurement_id, idx):
+        try:
+            measurement_dir = self.measurement_path.child(measurement_id)
+        except InsecurePath:
+            raise WebUIError(500, "Invalid measurement id")
+
+        measurements = measurement_dir.child("measurements.njson")
+
+        # This gets the line idx of the measurement file.
+        # XXX maybe implement some caching here
```
bassosimone: A possible way of implementing caching is that you cache the last openned file (or a subset of its lines)

bassosimone:
Well, then, I finished code review and did some light testing of everything. I am going to merge this large branch as soon as travis terminates such that we can push further fixes onto smaller and easier to review pull requests before finally reaching a stable release!

bassosimone:
Well, it seems I finally managed to stand up in front of this branch without being overwhelmed by complexity and this means that after a couple more fixes I'll definitely merge this.

14333070_10211217238653784_1566572335242110964_n

matcheydj:
Hi TheTorProject/ooni-p,

I'd like to add you to my professional network on LinkedIn.

- David

Confirm that you know David: https://www.linkedin.com/comm/start/accept-invitation?sharedKey=Xg4XQcbk&invitationId=6188766518159958016&trk=eml-guest-invite-cta&trkEmail=eml-invite_guest-null-2-null-null-0%7Eitucqbry%7E9e








You received an invitation to connect. LinkedIn will use your email address to make suggestions to our members in features like People You May Know. Unsubscribe here: https://www.linkedin.com/e/v2?e=0-itucqbry-9e&t=lun&midToken=AQFmrFVhGEd36Q&ek=invite_guest&loid=AQHx8lsXb5DX3AAAAVeLqub5ZlZ9euOpK4TsEOXJirqfqmW2YLFoF3hO5UYMRr9oSoYycCvLxf4nTv2tmhj91lBWlhwvKScTrTwj6CL0FZpyEkt0SrubHKPP9h1-Q5_VLQowiy0A4Bi3202y64YrblY1XStvvJAL9A7_S7tkIQaoPX2nInoCCUF0aZRmZub_fv8y&eid=0-itucqbry-9e


This email was sent to reply@reply.github.com.

If you need assistance or have questions, please contact LinkedIn Customer Service: https://www.linkedin.com/e/v2?e=0-itucqbry-9e&a=customerServiceUrl&ek=invite_guest

&copy; 2016 LinkedIn Corporation, 2029 Stierlin Court, Mountain View CA 94043. LinkedIn and the LinkedIn logo are registered trademarks of LinkedIn.

PR #588, opened 2016-08-17
SNI support for TLS handshake

darkk:
Sometimes https traffic is blocked on SNI basis, so I've written some code to test that.

It should be probably merged after 2.0.0 release or rebased onto 2.0.

### DIFF
--- a/ooni/nettest.py
+++ b/ooni/nettest.py
@@ -261,6 +261,7 @@ class NetTestLoader(object):
         into the usageOptions of the NetTestLoader.
         """
         if getattr(test_class.usageOptions, 'optParameters', None):
+            # the attribute may have `None` value, so getattr(o, name, []) can't be used
             for parameter in test_class.usageOptions.optParameters:
                 # XXX should look into if this is still necessary, seems like
                 # something left over from a bug in some nettest.
@@ -268,6 +269,9 @@ class NetTestLoader(object):
                 if len(parameter) == 5:
                     parameter.pop()
                 self.usageOptions.optParameters.append(parameter)
+
+        for flag in getattr(test_class.usageOptions, 'optFlags', []):
+            self.usageOptions.optFlags.append(flag)
 
         if getattr(test_class, 'inputFile', None):
             self.usageOptions.optParameters.append(test_class.inputFile)
--- a/ooni/nettests/experimental/tls_handshake.py
+++ b/ooni/nettests/experimental/tls_handshake.py
@@ -46,6 +46,8 @@ from ooni.utils import log
 from ooni.errors import InsufficientPrivileges
 from ooni.settings import config
 
+HAS_SNI = getattr(SSL.Connection, "set_tlsext_host_name", None) is not None
+
 ## For a way to obtain the current version of Firefox's default ciphersuite
 ## list, see https://trac.torproject.org/projects/tor/attachment/ticket/4744/
 ## and the attached file "get_mozilla_files.py".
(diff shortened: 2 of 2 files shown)


ooni/nettest.py
```diff
@@ -269,6 +270,9 @@ def _accumulateTestOptions(self, test_class):
                     parameter.pop()
                 self.usageOptions.optParameters.append(parameter)
 
+        for flag in getattr(test_class.usageOptions, 'optFlags', []):
+            self.usageOptions.optFlags.append(flag)
+
```
hellais: Yes good catch! I also noticed this while implementing the whatsapp test.

ooni/nettests/experimental/tls_handshake.py
```diff
@@ -126,9 +129,11 @@ class HandshakeOptions(usage.Options):
          'Remote host IP address (v4/v6) and port, i.e. "1.2.3.4:443"'],
         ['port', 'p', None,
          'Use this port for all hosts, regardless of port specified in file'],
+        ['sni', None, None, 'Use SNI for Server Name Indication field'],
         ['ciphersuite', 'c', None ,
          'File containing ciphersuite list, one per line'],]
     optFlags = [
+        ['no-sni', None, 'Disable SNI'],
```
hellais: I don't fully understand the purpose of this flag. It seems like the default behaviour is to **not** use SNI, so in which cases would this flag trigger a different behaviour.

I see you added a note below about this, perhaps we can either remove the flag or make the default behaviour be inferred depending on whether or not we have SNI support.

ooni/nettests/experimental/tls_handshake.py
```diff
@@ -761,7 +785,7 @@ def handshakeSucceeded(connection):
             ## (which would be visible in pcaps anyway, since the FQDN is
             ## never encrypted) I do not see a way for this to log any user or
             ## identifying information. Correct me if I'm wrong.
-            self.report['session_key'] = session_key
+            self.report['session_key'] = {'b64': base64.standard_b64encode(session_key)} # json can't do binary
```
hellais: It's probably good to use the same format we use for binary HTTP responses here. You can see how that is done here: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/common/http_utils.py#L35.

ooni/nettests/experimental/tls_handshake.py
```diff
@@ -743,6 +766,7 @@ def handshakeSucceeded(connection):
 
             self.report['host'] = host
             self.report['port'] = port
+            self.report['sni'] = self.sni
```
hellais: You probably want to just set this inside of the `setUp` method as it doesn't need to be changed across time. It's also more robust against possible failures of the test.

hellais:
Minor comments to this test. I will put the merge for this on hold until we get the 2.0.0 branch merged into master.

Good work 👍

ooni/nettests/experimental/tls_handshake.py
darkk: I don't know what's the best default option from technical standpoint:
  * no sni, contra: that's not what modern browser do
  * sni, contra: fail the test on old python OpenSSL library 
  * based on SNI support, contra: same command-line test launched on different boxes gives different results

ooni/nettests/experimental/tls_handshake.py
darkk: I was unsure if there was some reason to bind `host` and `port` so late, so I just droped `sni` flag close to the host/port pair. Are you aware of any reason to do that?

hellais: I think the reason is that it uses the `getpeername()` method of `SSL.Connection` and this could be different than what is specified as an input, I think.

It's a while that I haven't looked at this code and I am not sure if this is actually desired behaviour. I suspect it's probably wisest to log both in different keys.

ooni/nettests/experimental/tls_handshake.py
hellais: A couple of things:

* I think we want to make it robust to running in unsupported versions. When it doesn't do SNI due to an out of date version we could log this in the report so that we can take this into account when we assess the results.

* The value of SNI, if it's not explicitly specified, should be inferred from the value of `--host` or from the `input`.

hellais:
I think it would be cool to land this into master. It also has a useful fix to `NetTest` class that at minimum should make it's way into master.

Given the fact tls_handshake is experimental anyways, I don't see any harm in making experimental changes in it.

hellais:
Issue moved to [ooni/probe-legacy #10](https://github.com/ooni/probe-legacy/issues/10) via [**ZenHub**](https://www.zenhub.com/)

Issue #606, opened 2016-09-20
Missing country tests lists (2.0rc3)


kaerumy:
OONI Probe 2.0rc3  in ooni/resources there is citizenlab-test-lists, only the following files exist:

```
00-LEGEND-category_codes.csv
00-LEGEND-country_codes.csv
00-proposed-category_codes.csv
cis.csv
global.csv
```
web-full.yaml

<snip>
```
- name: Runs the Web Connectivity Test
  ooni:
    test_name: web_connectivity
    file: $citizenlab_${probe_cc}_urls
```
When run manually through web interface:

![image](https://cloud.githubusercontent.com/assets/1060743/18664483/5335eb24-7f55-11e6-8665-186e9f70b4e2.png)

It simply runs the web connectivity tests for global urls twice instead of detected local country list (MY).

hellais:
The way in which updates of inputs are managed has been changed in a way that reduced bandwidth consumption and allows us to also push updates to inputs more quickly, however this means that on first start ooniprobe does not have any resources or inputs.

What I suspect is going on here is that the first time the daily task to update the inputs runs, it still has not downloaded the geoip data files and therefore it's not able to determine the country code when doing the first fetch of the resources.

What will happen is that the next time the scheduled to update the inputs (every 24 hours) will run it will have the resources in there and so it will download the inputs for your country as well. This is to say that in 24 hours you should begin testing also the country specific list, but this is for sure a bug that should be solved.

I am going to add some extra logic to handle the first time running of the scheduler so that it will first download the resources for every country and then download the country specific ones.

kaerumy:
The country list was downloaded automatically, but when web connectivity tests were run by the scheduler it skipped the local list again.

From the file timestamp, it looks like the country list was downloaded/created 1 hours (8am local time) before the web connectivity tests (9am local time) were run.

But the no tests were run against the local cc list (MY)

hellais:
@kaerumy can you cat the scheduler file to see what is the actual timestamp of when the test was run?
Also perhaps mention the local time on the machine (the timestamps in the scheduler file are in UTC).

kaerumy:
scheduler timestamp:
```
2016-09-21T00:00:09Z
```
as localtime: 2016-09-21T08:00:09+08:00

Interestingly, the lock file was never removed despite the test via web UI saying it has completed, so this morning's test  at 0 UTC or  8:00 GMT+8 never ran

kaerumy:
[...]

PR #608, opened 2016-09-20
On first run of the UpdateInputsAndResources task fetch the resources…

hellais:
… with unknown country code

This fixes: https://github.com/TheTorProject/ooni-probe/issues/606


hellais:
@bassosimone says that I can merge. ⛪

Issue #602, opened 2016-09-13
Implement tests for instant messaging apps


hellais:
As the first iteration we should be implementing tests that check for the reachability of the endpoints used by IM apps when logging in and when performing activity on them.

In first pass I plan on implementing tests for the following apps based on the fact that they are the most popular ones and are fairly simple to implement:
- [x] Whatsapp 
- [x] Facebook Messenger
- [ ] Viber
- [ ] Telegram

This gist contains some user statistics for popular IM apps:
https://gist.github.com/hellais/19ae8f69a9772ffb177d68aef9e834be

PR #610, opened 2016-09-22
Feature/im tests

hellais:
This branch will contain pull requests for the following two IM tests:
- Facebook Messenger
- Whatsapp

Implementing https://github.com/TheTorProject/ooni-probe/issues/602


anadahz:
Minor improvements to whatsapp test:
It may worth to check if the IP list `https://www.whatsapp.com/cidr.txt` is accessible or blocked similar to the case of the whatapp's web interface check: https://github.com/TheTorProject/ooni-probe/pull/610/files#diff-796a6002beac082120abbbdb201639e1R320

hellais:
> It may worth to check if the IP list https://www.whatsapp.com/cidr.txt is accessible or blocked 

How is this relevant to assessing if whether or not the whatsapp application is working or not?

anadahz:
> How is this relevant to assessing if whether or not the whatsapp application is working or not?

A number of filtering implementations and censrors are using the CIDR to block the WhatApp application.
Knowing if the WhatsApp CIDR URL is being blocked could be an interesting measurement that can shed some light when the mobile application, the website, or the web application of WhatsApp are being blocked.

This is a nice to have feature but lower priority.

ooni/nettests/blocking/facebook_messenger.py
```diff
@@ -0,0 +1,159 @@
...
+        tcp_blocking = False
+        for key in FB_HOSTNAMES.keys():
+            if self.report['facebook-{0}-dns-consistent'.format(key)] == False:
+                dns_blocking = True
+                log.msg("{0} is BLOCKED due to DNS blocking".format(key))
+                continue
+
+            # XXX We ignore stun reachability as it requires UDP
+            if key == 'stun':
+                continue
```
darkk: BTW, `echo 'ooni-probe garbage' | nc -u stun.fbsbx.com 3478` gives `Bad request` response :)

ooni/nettests/blocking/whatsapp.py
```diff
@@ -0,0 +1,418 @@
...
+            self.report['tcp_connect'].append(result)
+
+        @d.addErrback
+        def eb(failure):
+            result['status']['success'] = False
+            result['status']['failure'] = failureToString(failure)
+            self.report['tcp_connect'].append(result)
+            return failure
+
+        return d
```
darkk: Should `test_connect_to_port` be shared across tests like dnsLookup & doRequest?

hellais: Yes this would be a good idea, ideally it would be an addition to the `TCPTest` template.

I think however this is out of scope for this current PR.

hellais: So I thought a bit more about this and also about measuring latencies and I think this would require a bit of thought to come up with a good spec for the data format used to implement it.

I wrote down some notes on this here: https://github.com/TheTorProject/ooni-probe/issues/689.

I would say we defer this for the moment to some bigger PR in the future as it will also require breaking the data format of past tests and probably implementing some translation code to old to new format.

I would put it under this: https://github.com/TheTorProject/ooni-spec/issues/74

Issue #296, opened 2014-03-14
Recursion limit problem in TaskManager


hellais:
The task manager is currently designed to recursively call `_fillSlots` that will then call `_run` that then again will call `_fillSlots` on success or failure. This means that when there are a lot of tasks failing very quickly it is very likely that the default python recursion limit will be overcome (1000).

To reproduce this bug you can try and run a test with a long invalid input for example http_requests:

`ooniprobe blocking/http_requests -f data/complete.deck`

Note that the fact that this test fails is correct, however it fails in a surprising manner:

```
Unhandled error in Deferred:
Unhandled Error
Traceback (most recent call last):
  File "/ooni-probe/ooni/managers.py", line 153, in _failed
    super(LinkedTaskManager, self)._failed(result, task)
  File "/ooni-probe/ooni/managers.py", line 44, in _failed
    task.done.errback(failure)
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/internet/defer.py", line 423, in errback
    self._startRunCallbacks(fail)
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/internet/defer.py", line 490, in _startRunCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/internet/defer.py", line 577, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/ooni-probe/ooni/director.py", line 188, in measurementFailed
    log.msg("Failed doing measurement: %s" % measurement)
  File "/ooni-probe/ooni/utils/log.py", line 62, in msg
    print "%s" % msg
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/python/log.py", line 505, in write
    msg(message, printed=1, isError=self.isError)
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/python/threadable.py", line 53, in sync
    return function(self, *args, **kwargs)
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/python/log.py", line 185, in msg
    actualEventDict = (context.get(ILogContext) or {}).copy()
  File "/.virtualenvs/ooni-probe/lib/python2.7/site-packages/twisted/python/context.py", line 121, in getContext
    return self.currentContext().getContext(key, default)
exceptions.RuntimeError: maximum recursion depth exceeded
```

[...]

PR #617, opened 2016-10-07
Fix/robust scheduling

hellais:
Opening this pull request to apply to some comments on the PR.

Great work on this!

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@ $setup_ooniprobe = <<SCRIPT
 echo "deb http://deb.torproject.org/torproject.org jessie main" >> /etc/apt/sources.list
 
 # Install deb.torproject.org key
-gpg --keyserver keys.gnupg.net --recv 886DDD89
+gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
 gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | apt-key add -
 
 apt-get update
@@ -30,7 +30,7 @@ echo "Installing Tor..."
 
 echo "deb http://deb.torproject.org/torproject.org jessie main" >> /etc/apt/sources.list
 
-gpg --keyserver keys.gnupg.net --recv 886DDD89
+gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
 gpg --export A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89 | apt-key add -
 
 apt-get update
--- a/ooni/nettest.py
+++ b/ooni/nettest.py
@@ -6,7 +6,7 @@ import sys
 from twisted.internet import defer
 from twisted.python.filepath import FilePath
 from twisted.trial.runner import filenameToModule
-from twisted.python import usage, reflect
+from twisted.python import failure, usage, reflect
 
 from ooni import __version__ as ooniprobe_version, errors
 from ooni import otime
@@ -578,6 +578,9 @@ class NetTest(object):
         """
         This is a generator that yields measurements and registers the
         callbacks for when a measurement is successful or has failed.
+
+        FIXME: If this generator throws exception TaskManager scheduler is
+        irreversibly damaged.
         """
 
         for test_class, test_methods in self.testCases:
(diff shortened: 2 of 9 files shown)


ooni/nettest.py
```diff
@@ -591,11 +594,16 @@ def generateMeasurements(self):
                 test_instance._setUp()
                 test_instance.summary = self.summary
                 for method in test_methods:
+                    try:
+                        measurement = self.makeMeasurement(
+                            test_instance,
+                            method,
+                            test_input)
+                    except Exception:
+                        log.exception(failure.Failure())
```
hellais: I think it's better if you pass to the log.exception the actual exception so that we log also the exception type together with the traceback.

ex.

```
except Exception as exc:
    log.exception(exc)
```

darkk: Traceback will be lost in this case. `Failure()` takes current exception + trackback and passes it to the logger.

darkk:
That's still WIP.

ooni/nettest.py
hellais: Gosh, you are right. It does forget the traceback when you pass the exception explicitly. I do see however that `twisted.python.log` internally will create an empty failure by default when you pass it `None`, so I guess if we are to go this way it's enough to just call `log.exception()`

ooni/nettest.py
```diff
@@ -591,11 +594,16 @@ def generateMeasurements(self):
                 test_instance._setUp()
                 test_instance.summary = self.summary
                 for method in test_methods:
+                    try:
+                        measurement = self.makeMeasurement(
```
hellais: So thinking a bit more about this, what is not nice is that the constructor for `Measurement` does a bunch of things, including calling `setUp` that is a method that can actually throw an exception (https://github.com/TheTorProject/ooni-probe/blob/99367d6b724b69897215230fa58dc398eabc42db/ooni/tasks.py#L114).

I guess the cleaner way of managing this would be to move the setup of the NetTestCase instance inside of another method that is called inside of `makeMeasurement` and handle exceptions in there.

Vagrantfile
```diff
@@ -5,7 +5,7 @@ $setup_ooniprobe = <<SCRIPT
 echo "deb http://deb.torproject.org/torproject.org jessie main" >> /etc/apt/sources.list
 
 # Install deb.torproject.org key
-gpg --keyserver keys.gnupg.net --recv 886DDD89
+gpg --keyserver keys.gnupg.net --recv A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
```
bassosimone: @darkk good to know that keys should always be expressed in full

ooni/tests/test_scheduler.py
```diff
@@ -51,8 +52,40 @@ def task(subself):
...
+                with open(counter, 'w+') as fd:
+                    data = fd.read()
+                    fd.seek(0)
+                    if not data:
+                        fd.write('1')
+                        yield task.deferLater(reactor, 1, lambda: 42)
+                    else:
+                        # should not write anything as followup tasks should not be run
+                        fd.write(str(int(data) + 1))
+                        yield 34
```
hellais: This generator should be yielding a deferred and not an int. Otherwise it will trigger some exception.

darkk: That's true, but this code path should **not** be run at all :)

ooni/tests/test_scheduler.py
```diff
@@ -51,8 +52,40 @@ def task(subself):
...
+        identifier = "dummy"
+        dummy_st = DummyST(schedule='@daily', identifier=identifier, scheduler_directory=scheduler_directory)
+        dl = defer.DeferredList([dummy_st.run() for i in xrange(lockno)], consumeErrors=True)
+        @dl.addBoth
+        def so_what(results):
+            self.assertEqual(results[0], (True, None)) # wtf is `None` ?
```
darkk: It looks like deadlock :) It's
```
[Failure instance: Traceback (failure with no frames): <class 'twisted.internet.defer.TimeoutError'>: <ooni.tests.test_scheduler.TestScheduler testMethod=test_thundering_herd> (test_thundering_herd) still running at 120.0 secs
]
```

Try running the testcase with `env TTH_LOCKNO=10` and `env TTH_LOCKNO=1000` -- different number of waiters produces different results. IMHO, that's exactly the reason of the deadlock I've seen at least couple of times — the lockfile had `mtime` equal to the moment when the long-running measurement was done and bunch of ~100 accumulated tasks waiting in a queue was started.

darkk:
See also #296

ooni/nettest.py
darkk: I'd rather be explicit about failure creation. `log.exception(None)` does not look like _write current exception information_.

Moreover, it make some sense to occasionally review all `log.exception` calls to check if lost traceback is ok for this case.

ooni/nettest.py
darkk: I placed exception handling here instead of `makeMeasurement` (or it's replacement) to keep makeMeasurement code clean from logging. It's possible to skip bad measurement returning `None`, do you suggest it?

IMHO, the best way is to make the scheduler code exception-neutral instead of fixing exceptions one by one, I've attempted that and that was leading to quite significant refactoring that was probably to invasive for RC stage.

hellais:
I did a bit more testing on this and pushed some unittests for the full scheduler workflow. I am going to proceed with merging this branch.

Excellent work @darkk 👍

PR #620, opened 2016-10-07
The correct name for cloudfront based backends is "cloudfront"

hellais:


hellais:
I am going to proceed in merging this as it's a hotfix.

Issue #619, opened 2016-10-07
Investigate occasional locking of the web UI


hellais:
In testing the web UI it seems like sometimes the web UI becomes irresponsive for 2-3 seconds (HTTP requests stay pending for many seconds).

This is most probably due to blocking code that is interleaved with non blocking code.

We should look into places where there is code that is blocking either on IO or CPU and refactor it to use non blocking IO or spawn off threads.

Noticed in https://github.com/TheTorProject/ooni-probe/issues/612

darkk:
Have you found any root cause so far? I can try to get a reproducible test-case.

hellais:
@darkk #621 resolved one of the issues I was having. There may be more issues like this lurking around so I haven't closed it, but I was able to reproduce the problem before and after #621 (manually though, without a proper unittest as it requires browser interaction).

darkk:
I've checked pcap's & devtools — it's not the some sort of blocking code (confirmed by strace) or CPU-intensive code (confirmed by cpuacct.stat) that is causing 7…20 seconds lags, it's my Firefox-49.0 being slow crunching megabytes of javascript code in single process.

I can't reproduce the issue event **without** #621 fix using either Firefox instance with clear **profile** (Private browsing does not help) or chromium 53.0.2785.143.

It's quite interesting that ooni-probe UI makes my "day-to-day" FF instance to hog CPU and spend 50% of time running GC despite having 2G of free RAM, but that's probably is not worth debugging unless someone else complains. :-)

hellais:
This is fixed.

PR #621, opened 2016-10-07
When generating a summary do so inside of a separate thread to avoid …

hellais:
…blocking the reactor thread

This is related to https://github.com/TheTorProject/ooni-probe/issues/619.

I am going to use this branch to push other fixes related to this.

I believe this was the issue I was noticing more prominently when using the web ui, but there may also be other cases of this happening.

### DIFF
--- a/ooni/measurements.py
+++ b/ooni/measurements.py
@@ -1,5 +1,7 @@
 import json
 
+from twisted.internet import defer
+from twisted.internet.threads import deferToThread
 from twisted.python.filepath import FilePath
 from ooni.utils import log, is_process_running
 from ooni.utils.files import directory_usage
@@ -56,6 +58,7 @@ def generate_summary(input_file, output_file):
 
     with open(output_file, "w") as fw:
         json.dump(results, fw)
+    return results
 
 class MeasurementNotFound(Exception):
     pass
@@ -103,21 +106,27 @@ def get_measurement(measurement_id, compute_size=False):
 
 
 def get_summary(measurement_id):
+    """
+    Returns a deferred that will fire with the content of the summary
+     or will errback with MeasurementInProgress if the measurement has not
+     yet finished running.
+    """
     measurement_path = FilePath(config.measurements_directory)
     measurement = measurement_path.child(measurement_id)
 
     if measurement.child("measurements.njson.progress").exists():
-        raise MeasurementInProgress
+        return defer.fail(MeasurementInProgress)
 
     summary = measurement.child("summary.json")
     if not summary.exists():
-        generate_summary(
+        return deferToThread(
+            generate_summary,
             measurement.child("measurements.njson").path,
             summary.path
         )
 
     with summary.open("r") as f:
-        return json.load(f)
+        return defer.succeed(json.load(f))
 
 
 def list_measurements(compute_size=False):
--- a/ooni/ui/web/server.py
+++ b/ooni/ui/web/server.py
@@ -527,7 +527,9 @@ class WebUIAPI(object):
     @app.route('/api/measurement/<string:measurement_id>', methods=["GET"])
     @xsrf_protect(check=False)
     @requires_true(attrs=['_is_initialized'])
+    @defer.inlineCallbacks
     def api_measurement_summary(self, request, measurement_id):
+        log.warn("SUMMARY")
         try:
             measurement = get_measurement(measurement_id)
         except InsecurePath:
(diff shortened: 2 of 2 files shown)


darkk:
Do you have **significant** load latency reduction?

I mean, I don't know how does this code change anything due to GIL. Maybe `json` releases it, maybe it does not, maybe it does not matter at all (due to some other reason). It [does not change anything for me](https://github.com/TheTorProject/ooni-probe/issues/619#issuecomment-253193607), but if it improves any use-case you observe, it looks good for me.

I've briefly looked at `get_summary`, I see no obvious missing locks.

bassosimone:
@darkk as I understand it, but I may possibly be wrong, the specific issue should be that an operating doing some I/O and some CPU is started on the twisted reactor's thread while there are open connections waiting to be served. This is something that could stall the web user interface and doing operations in a background thread should make sure that the task of responding to the web ui is more responsive.

Now, that said, I believe there is another reason why the user interface could appear frozen and also that perhaps is worth investigating. My experience with Neubot is that, when you receive lots of data from the backend, and the browser loops through it, the web interface is stalled because the single JavaScript thread is spending too much time to process the data. A way to avoid that would be to process data in small batches and then reschedule the work by calling `setTimeout(function () { /* continue iterating over data */ }, 0.0)` so to give the browser time to breathe.

A third possible issue is that, if too much data at a batch is received, this could consume too much memory and/or resources. I don't remember well how that part of ooni works but in general I assume it would make sense to have either paginated APIs (like for example Twitter does) so to process data in smaller batches or perhaps to directly switch to use WebSockets (do people still use them or are there better transport?) and make sure that you send data line by line.

Does it make sense?

hellais:
The problem manifests itself if you have a very big report and you navigate to the measurement page for that particularly large report for the first time.

By particularly large I mean something like 100 MB.

When doing so the `get_summary` method is called that will perform the reading (in blocking IO style) and JSON parsing inside of the main twisted reactor event loop. This means that the event loop will be occupied by the blocking IO reads and the CPU intensive JSON parsing operations and will therefore not be able to process the other pending HTTP requests.

It's not a huge issue, the UI becomes responsive again in about 5 seconds, but it is something noticeable for me.

bassosimone:
(Regarding `json` releasing the lock, a cursory look a simplejson/simplejson _speedups.c does not show stuff like `PyGILState_STATE state = PyGILState_Ensure();` and searching for `gil` within that file yields no results. Either is a macro or something more elaborate or diverse from what I expect, or it does not release the GIL.)

hellais:
> (Regarding json releasing the lock, a cursory look a simplejson/simplejson _speedups.c does not show stuff like PyGILState_STATE state = PyGILState_Ensure(); and searching for gil within that file yields no results. Either is a macro or something more elaborate or diverse from what I expect, or it does not release the GIL.)

note that we don't use simplejson, but rather the default json module which is pure python and GIL related trickery is happening there.

bassosimone:
> note that we don't use simplejson, but rather the default json module which is pure python

True, but [the json in standard library is simplejson](https://github.com/simplejson/simplejson/blob/c80be5d00fe278cef3d76b6d68b8918ed849c99b/README.rst): "simplejson is the externally maintained development version of the json library included with Python 2.6 and Python 3.0, but maintains backwards compatibility with Python 2.5."

However, looking in my local install, I do not see speedups.

> and GIL related trickery is happening there.

Ah, interesting, I didn't know that! What kind of GIL trickery they do in their pure Python?

hellais:
> Ah, interesting, I didn't know that! What kind of GIL trickery they do in their pure Python?

ehm, typed too fast. I meant to say **NO** GIL trickery is happening.

darkk:
> Does it make sense?

That's perfect reasoning. The only thing I wanted to tell, that my complains about bad wui performance were actually caused by 935 background tabs opened in Firefox. So these complains were bogus. @hellais means different test-case.

> blocking IO reads and the CPU intensive JSON

Yep, I've simulated `get_summary` to get some baseline with 
```
python -c 'json.loads(json.dumps([json.loads(i) for i in open("measurements/20161007T000036Z-RU-AS8997-web_connectivity/measurements.njson")]))'
```
It takes ~0.3s of kernel time and ~3.0s of user time. So, yeah, it makes some sense to 

> default json module which is pure python
Not so pure:
```
>>> json.scanner.make_scanner == json.scanner.c_make_scanner
True
>>> json.scanner.make_scanner == json.scanner.py_make_scanner
False
```
But I see no GIL tricks in _json.so with `nm -D`

hellais:
So do we think this is good to merged? Anything reason not to merge or some alternative solution to this?

darkk:
LGTM, as latency improvement is confirmed.

PR #622, opened 2016-10-11
Make no intersection amongst decks

hellais:
* Create 2 new decks that are currently quite weak, but can be expanded


hellais:
I did this thing of created two new "filler" decks for IM and tor, so that when we have more complete tests for instant messaging apps and tor we can push this to users that have opted in to testing these categories of apps.

data/decks/web.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-name: Web test deck without HTTP Invalid Request Line
-description: This deck runs HTTP Header Field Manipulation, and the Web Connectivity test
+name: Web test deck
+description: This deck runs HTTP Header Field Manipulation and the Web Connectivity test
```
darkk: By the way, is `${probe_cc}` cached for a long time? I mean, how will `ooni-probe` behave if the laptop is suspended in one country and awaken in another?

hellais: We cache geoip lookups for up to 10 minutes, so unless a user is travelling very fast between different locations this should not create issues.

hellais:
Got the LGTM. Mergin'.

PR #628, opened 2016-10-12
Fix/pidfile failing unlink

darkk:
Making twistd a bit more happy, avoiding alike exceptions written to `ooniprobe.log` during `ooniprobe-agent stop`:
 
```
2016-10-11 11:11:47,930311+0200 [-] Received SIGTERM, shutting down.
2016-10-11 11:11:47,931315+0200 [-] Main loop terminated.
2016-10-11 11:11:47,931542+0200 [-] Failed to unlink PID file:
        Failure: exceptions.OSError: [Errno 2] No such file or directory: '/var/lib/ooni/twistd.pid'
        
2016-10-11 11:11:47,931727+0200 [-] Server Shut Down.
```

### DIFF
--- a/ooni/scripts/ooniprobe_agent.py
+++ b/ooni/scripts/ooniprobe_agent.py
@@ -121,6 +121,17 @@ def get_running_pidfile():
             return pidfile
     raise NotRunning
 
+def is_stale_pidfile(pidfile):
+    try:
+        with open(pidfile) as fd:
+            pid = int(fd.read())
+    except Exception:
+        return False # that's either garbage in the pid-file or a race
+    return not is_process_running(pid)
+
+def get_stale_pidfiles():
+    return [f for f in [config.system_pid_path, config.user_pid_path] if is_stale_pidfile(f)]
+
 def status_agent():
     try:
         get_running_pidfile()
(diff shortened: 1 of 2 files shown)


ooni/scripts/ooniprobe_agent.py
```diff
@@ -121,6 +121,16 @@ def get_running_pidfile():
             return pidfile
     raise NotRunning
 
+def is_stale_pidfile(pidfile):
+    if os.path.exists(pidfile):
+        with open(pidfile) as fd:
```
hellais: I would wrap this inside of a try except to avoid races between the `exists` check and the opening of the file.

ooni/scripts/ooniprobe_agent.py
```diff
@@ -121,6 +121,16 @@ def get_running_pidfile():
...
+            pid = int(fd.read())
+            return not is_process_running(pid)
+    return False
+
+def get_stale_pidfiles():
+    return [_ for _ in [config.system_pid_path, config.user_pid_path] if is_stale_pidfile(_)]
```
hellais: Minor nitpick: generally the `_` variable is used for things you don't care about and want to just ignore. In this case I would use any single letter variable of your choosing (`f`?).

hellais:
Two minor nitpicks, that once fixed is good to merge.

ooni/scripts/ooniprobe_agent.py
darkk: Yep, that's true. OTOH, I can't easily imagine the race to trigger that.

hellais:
LGTM!

Issue #630, opened 2016-10-12
smear load generated by scheduler


darkk:
New `croniter`-based scheduler ticks in UTC and runs `@daily` tasks at 0:00.

That may hammer low-traffic websites in lists as the lists are not shuffled on load.

Load smearing should satisfy following requirements:
- [x] prompt run after first installation
- [x] survive reboot & daemon restart in a sane way
- [x] be aware of the fact that the probe is not running 24/7

hellais:
Issue moved to [ooni/probe-legacy #27](https://github.com/ooni/probe-legacy/issues/27) via [**ZenHub**](https://www.zenhub.com/)

PR #631, opened 2016-10-12
Smear load generated by scheduler, see #630

darkk:
- [x] It runs ASAP after the installation.
- [x] It survives reboot in predictable way: avoids double-run and does not skip scheduler cycle on restart.
- [x] It supports non-24/7 operations as it targets running @daily task within 2.4 hours window, so the probe that was down will likely run a task on boot and boot time is random enough for load smearing purposes.

xref: #630


ooni/agent/scheduler.py
```diff
@@ -82,18 +83,28 @@ def __init__(self, schedule=None, identifier=None,
...
         )
+        self._smear_coef = random.random()
 
     def cancel(self):
+        if not self._last_run_lock.locked:
+            # _last_run_lock.release() will throw if we try to release it
+            log.err('BUG: cancelling non-locked task {} without holding lock'.format(self.identifier))
+            return
+        # probably, cancelling the task TAKEN the lock is even worse :-)
         self._last_run_lock.release()
```
darkk: I added that to be able to `unschedule()` the task. But...

WAT?
====

ooni/contrib/croniter.py
```diff
@@ -239,7 +239,7 @@ def _get_next(self, ret_type=None, is_prev=False):
             result = self._calc(self.cur, expanded, is_prev)
         self.cur = result
 
-        if ret_type == datetime.datetime:
+        if issubclass(ret_type, datetime.datetime):
```
darkk: - [ ] upstream it

ooni/tests/test_scheduler.py
```diff
@@ -274,43 +274,97 @@ def test_deck_run_twice(self, mock_resources, mock_probe_ip,
...
+            # nothing happens during an hour
+            dummy_clock.advance(60 * 60 - 46)
+            self.assertEqual(FakeDatetime.utcnow(), datetime(2000,1,2, 7,59,59))
             for t in scheduler_service._scheduled_tasks:
-                with open(os.path.join(self.scheduler_directory,
-                                       t.identifier), 'w') as out_file:
-                    yesterday = (now_time - timedelta(days=1,
-                                                      hours=2)).strftime(DT_FRMT)
-                    out_file.write(yesterday)
-            dummy_clock.advance(30)
```
darkk: @hellais did I got right idea of the scheduler file-clock ticking backward?

ooni/tests/test_scheduler.py
```diff
@@ -274,43 +274,97 @@ def test_deck_run_twice(self, mock_resources, mock_probe_ip,
...
+                    self.assertLessEqual(t._smear_coef, 1.0)
+                    t._smear_coef = random.random()
+            self.assertEqual(len(launches), len(scheduler_service._scheduled_tasks))
+            self.assertEqual({k: len(v) for k, v in launches.iteritems()}, dict.fromkeys(launches.iterkeys(), 2))
 
             # We check that the run method of the deck was called twice
+            # NB: That does NOT check that @daily task was called exactly twice
             self.mock_deck.run.assert_has_calls([
                 mock.call(mock_director), mock.call(mock_director)
             ])
```
darkk: Does it really behave as expected?

darkk:
Side-effect: `update-inputs` may be scheduled **later** than `run-deck`. IMHO, that's not critical.

hellais:
If I understand this correctly, basically we are offsetting ALL ScheduledTasks by some value in the range [0, $period * 0.1].
The value by which we offset it is going to be fixed until the ooniprobe-agent is restarted.

Also this smear factor applies to run-deck tasks, but also to the rest of the tasks.

I think it would be better if:

* [ ] We had the ability to choose which scheduled tasks should be affected by the smear factor and which not. Some tasks we care to have offset by some value, but other we actually don't care and some we may actually want them to be scheduled precisely (for example those not doing network activity).

* [ ] The smear factor is recomputed after a successful execution.

darkk:
> we may actually want them to be scheduled precisely

Do you have an example why we may actually want that? 

> factor is recomputed after a successful execution.

Yep, I can imagine example when it matters: the probe is running 24/7, ISP enforces reconnection every 24 hours (precisely) — we may be unlucky doing same measurements at the moment of reconnection on an on.

hellais:
> Do you have an example why we may actually want that?

In looking at the current scheduled tasks we have, there is actually no real need to have them be triggered with an accurate timing. Given that, I think we should perhaps add support for this when we have some tasks that do require that.

(Thanks for reality checking my requirements, I have the tendency of not being as KISS as I should).

> Yep, I can imagine example when it matters: the probe is running 24/7, ISP enforces reconnection every 24 hours (precisely) — we may be unlucky doing same measurements at the moment of reconnection on an on.

Indeed. Or the time at which the probe is scheduled to run, because we are unlucky, is a time in which the users network is under a considerable amount of load for other reasons and we end up doing measurements in a time where there is a certain amount of load.

An easy way of implementing this would be adding another hook after a successful task execution that recomputes the smear factor value (https://github.com/TheTorProject/ooni-probe/blob/master/ooni/agent/scheduler.py#L139).

ooni/tests/test_scheduler.py
hellais: yes you are right. This is just asserting that is was "called **at least** twice", but not exactly twice.

We can enforce exactly twice by doing below:

```
self.assertEqual(self.mock_deck.run.call_count, 2)
```

ooni/tests/test_scheduler.py
```diff
@@ -274,43 +274,97 @@ def test_deck_run_twice(self, mock_resources, mock_probe_ip,
...
+                        self.assertRegexpMatches(last_run, '^2000-01-02T08:0.:[03]0Z$')
+                        if last_run == '2000-01-02T08:00:00Z':
+                            zero += 1
+            self.assertGreater(hourly, 0)
+            self.assertGreater(daily, 0)
+            self.assertLess(zero, hourly)
+            self.assertLessEqual(zero, 1) # should ALMOST never happen
+
+            # leaping to the end of the day
+            dummy_clock.advance((datetime(2000,01,02,23,59,59) - FakeDatetime.utcnow()).total_seconds())
```
hellais: Be careful when prefixing numbers with a leading zero as python2 and python3 treat this differently, but in python2 this means that the number is expressed in base8, while in python3 it's a syntax error.

```
$ python2 -c 'print("%s" % 010)'
8
$ python3 -c 'print("%s" % 010)'
  File "<string>", line 1
    print("%s" % 010)
                   ^
SyntaxError: invalid token
```

darkk: Thanks, I was unaware of this py3 syntax change.

ooni/agent/scheduler.py
hellais: Yes, this logic here is very fragile. You are right to add this extra check. 

This was created within the context of being able to reload the list of decks that are currently enabled (`refresh_deck_list`). We need in there to be able to unschedule decks that have been disabled by a user.

Other issues with this method are:

1. If a task is currently running, it will continue execution and not interrupt while it's running

2. Because of 1. if a task was unscheduled and then rescheduled while another one was running, you will end up with two tasks running at the same time. If you toggle a certain task many times this can lead you having a bunch of tasks running at the same time

There are possibly also other issues that I haven't yet thought of.

ooni/contrib/croniter.py
hellais: xref: https://github.com/taichino/croniter/pull/70

hellais: The PR for that has been merged and landed in a new release. We should update our vendorred version of it to be in sync with their latest release: https://github.com/taichino/croniter/blob/master/docs/CHANGES.rst#0313-2016-11-01

hellais:
I am thinking of making this land into 2.2.0. @darkk do you see any issues with merging this?

ooni/tests/test_scheduler.py
hellais: Update: this probably should be changed to avoid issues in the future. Since 1 < 8 & 2 <8 it's by chance equivalent, but we may have issues because, for example, `09` is an invalid base 8 number.

ooni/tests/test_scheduler.py
```diff
@@ -274,43 +274,97 @@ def test_deck_run_twice(self, mock_resources, mock_probe_ip,
...
+            # ticking through six hours
+            dummy_clock.pump([random.uniform(0, 120) for i in xrange(6*60)])
+            for t in scheduler_service._scheduled_tasks:
+                with open(os.path.join(self.scheduler_directory, t.identifier)) as in_file:
+                    # randomized clock kills 30s resolution of the scheduler
+                    self.assertRegexpMatches(in_file.read(), '^2000-01-03T0[012]:..:..Z$')
+            self.assertGreater(FakeDatetime.utcnow(), datetime(2000,1,3, 5,0,0)) # should be ~6:00
+
+            # verify, that double-run does not happen even in case of reseeding (reboot/restart)
+            dummy_clock.advance((datetime(2000,01,03,23,59,59) - FakeDatetime.utcnow()).total_seconds())
```
hellais: Same as above.

hellais:
I am going to proceed with merging this into master so we can test it on staging.

darkk:
Yep, so far I've not recalled what was wrong with that code.
Anyway, it should be better than triggering everything at `00:00 UTC` precisely.

PR #632, opened 2016-10-14
Annotate reports with the platform on which they are running.

hellais:


ooni/settings.py
```diff
@@ -325,6 +326,20 @@ def web_ui_url(self):
...
+    def platform(self):
+        if os.path.exists('/etc/default/lepidopter'):
+            return 'lepidopter'
+        system = platform.system()
+        if system == 'Darwin':
+            return 'macos'
+        elif system == 'Linux':
+            return 'linux'
+        elif system == 'Windows':
+            # Really?
```
bassosimone: :-D

bassosimone:
Looks good to me!

bassosimone:
Yes, I think this is a very good thing to do! How would this end up into the report? Just to know it and do something equivalent for measurement-kit and possibly also at least add an issue to specify this information into the specification of our data format.

ooni/settings.py
```diff
@@ -325,6 +326,20 @@ def web_ui_url(self):
...
+            return 'windows'
+        return 'unknown'
```
darkk: Why don't you want to return `platform.system().lower()` instead of `'unknown'`?

PR #633, opened 2016-10-14
Ensure annotations are always dicts

hellais:
Damn, hastily pushed and merged without actually waiting for unittest to finish.

### DIFF
--- a/ooni/nettest.py
+++ b/ooni/nettest.py
@@ -169,10 +169,12 @@ class NetTestLoader(object):
     requiresTor = False
 
     def __init__(self, options, test_file=None, test_string=None,
-                 annotations={}):
+                 annotations=None):
         self.options = options
-        # Ensure annotations are always dicts
+        if annotations is None:
+            annotations = {}
         if not isinstance(annotations, dict):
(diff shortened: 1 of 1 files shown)


ooni/nettest.py
```diff
@@ -171,6 +171,9 @@ class NetTestLoader(object):
     def __init__(self, options, test_file=None, test_string=None,
                  annotations={}):
         self.options = options
+        # Ensure annotations are always dicts
+        if not isinstance(annotations, dict):
+            annotations = {}
```
darkk: There are actually two catches here

- [ ] using mutable object as a default argument
```
class Foo(object):
    def __init__(self, annotations={}):
        self.annotations = annotations
        self.annotations['foo'] = self.annotations.get('foo', 0) + 1
        print self.annotations
a, b, c = Foo(), Foo(), Foo()
```

- [ ] dropping the data silently. IMHO, it should be something like `if ann is None: ann = {}; if not isinstance(ann, dict): throw TypeError`

darkk:
LGTM. I'm unsure of `repr(annotation)` is worth logging.

PR #637, opened 2016-10-16
Explicitly test all Twisted versions, add pip list, format template

anadahz:
Successfully debugging an ooniprobe build/run implies that we know the
specific package and version dependency of a build.
* Add missing Twisted versions from travis environment
* Add the output of pip list to the travis build log
* Minor fixes to irc template format


hellais:
If you think a branch is ready for review remember to put the "ready for review" tag.

Looking at this I think it's ready for review and good to merge. Thanks ! 👍

PR #655, opened 2016-10-26
Fix/paths

hellais:
This also includes a fix that removes some invalid metadata from the setup.py

### DIFF
--- a/ooni/settings.py
+++ b/ooni/settings.py
@@ -156,6 +156,9 @@ if hasattr(sys, 'real_prefix'):
 
 # These are the the embedded settings
 _SETTINGS_INI = os.path.join(OONIPROBE_ROOT, 'settings.ini')
+# This can be overridden by an environment variable for end to end testing
+if 'OONIPROBE_SETTINGS' in os.environ:
+    _SETTINGS_INI = os.environ['OONIPROBE_SETTINGS']
 
 USR_SHARE_PATH = '/usr/share/ooni'
 VAR_LIB_PATH = '/var/lib/ooni'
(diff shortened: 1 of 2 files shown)


ooni/settings.py
```diff
@@ -156,6 +156,9 @@
 
 # These are the the embedded settings
 _SETTINGS_INI = os.path.join(OONIPROBE_ROOT, 'settings.ini')
+# This can be overridden by an environment variable for end to end testing
+if 'OONIPROBE_SETTINGS' in os.environ:
+    _SETTINGS_INI = os.environ['OONIPROBE_SETTINGS']
```
darkk: `os.environ.get('OONIPROBE_SETTINGS', os.path.join(OONIPROBE_ROOT, 'settings.ini'))` is a bit more pythonic :)

setup.py
```diff
@@ -110,7 +110,6 @@
 Intended Audience :: Information Technology
 Intended Audience :: Science/Research
 Intended Audience :: Telecommunications Industry
-License :: OSI Approved :: BSD Licens
```
darkk: Was it enough to add `e` to `Licens` to make it valid?

ooni/settings.py
hellais: Yes that is true, but I prefer to have it be explicit and clear that this is what is going on here. Moreover we may want to override other settings via environment variables and it's good to have a consistent pattern for doing so.

setup.py
hellais: Yeah that's true. I can fix it that way too.

darkk:
LGTM

Issue #650, opened 2016-10-24
ooniprobe stops after configuration


Te-k:
Hi,
Minor bug : the first time I launch ooniprobe (v2.0.0), the scripts asks for configuration information and stops which it counter-intuitive as I wanted to run a test. I had to relaunch ooniprobe again. It would be better to launch the test directly after configuration.

Logs after clean installation (no .ooni directory, ooni installed through pip): 
```
$ oonideckgen 
Checking for update of resources
Deck written to /home/user/projects/ooni/web-full.yaml
Run ooniprobe like so:
ooniprobe -i /home/user/projects/ooni/web-full.yaml
$ ooniprobe -i web-full.yaml 
It looks like this is the first time you are running ooniprobe
Please take a minute to read through the informed consent documentation and understand what are the risks associated with running ooniprobe.
Press enter to continue...

Type "yes" if you are fully aware of the risks associated with using ooniprobe and you wish to proceed
> yes

Now help us configure some things!
Should we upload measurements to a collector? (Y/n) y
Should we include your IP in measurements? (y/N) n
Should we include your ASN (your network) in measurements? (Y/n) y
Should we include your Country in measurements? (Y/n) y
How would you like reports to be uploaded? (onion, https, cloudfront) onion
$
```

Version information  : 
  * Ubuntu 16.04.1
  * python 2.7.12
  * ooniprobe version: 2.0.0

hellais:
Yes I agree this behaviour is not ideal. I am going to push a fix to this shortly.

Thanks for reporting this!

PR #658, opened 2016-10-27
When ooniprobe is run for the first time. Initialize it and then run …

hellais:
…the command the user wanted.

This fixes: https://github.com/TheTorProject/ooni-probe/issues/650

### DIFF
--- a/ooni/ui/cli.py
+++ b/ooni/ui/cli.py
@@ -171,6 +171,10 @@ def director_startup_other_failures(failure):
 
 def initializeOoniprobe(global_options):
     print("It looks like this is the first time you are running ooniprobe")
+    if not sys.stdin.isatty():
+        print("ERROR: STDIN is not attached to a tty. Quiting.")
+        sys.exit(8)
+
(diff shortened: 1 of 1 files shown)


ooni/ui/cli.py
```diff
@@ -224,25 +224,23 @@ def initializeOoniprobe(global_options):
...
+    print("ooniprobe is now initialized. You can begin using it!")
 
 def setupGlobalOptions(logging, start_tor, check_incoherences):
     global_options = parseOptions()
 
     config.global_options = global_options
 
-    if not config.is_initialized():
-        log.err("You first need to agree to the informed consent and setup "
-                "ooniprobe to run it.")
```
darkk: (sidenote) It's interesting that the message was not actually displayed to user.

LGTM

hellais: Yeah that is due to how the logging is setup and it "stealing" stdin and stdout.

I think we will have to at some point have a better separation between messages that are "logs" and messages that are meant for the user.

PR #660, opened 2016-10-28
Pass the missing start_tor argument to director start command

### LINKED ISSUE #636: vanilla_tor test is not working when run from the command line
In ooniprobe 2.0.0 when running the vanilla_tor test from the command the following expection is raised:

```
$ ooniprobe vanilla_tor
No test deck detected
Tor is not running. Skipping IP lookup via Tor.
Looking up your IP address via ubuntu
Running task update-inputs
Did not run update-inputs
Looking up collector and test helpers with http://nkvphnp3p6agi5qq.onion
Querying backend http://nkvphnp3p6agi5qq.onion/bouncer/net-tests with {'net-tests': [{'test-helpers': [], 'version': '0.1.0', 'name': 'vanilla_tor', 'input-hashes': []}]}
Unhandled Error
  File "/usr/local/lib/python2.7/dist-packages/ooniprobe-2.0.0-py2.7.egg/ooni/scripts/ooniprobe.py", line 34, in run
    task.react(ooniprobe)
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/task.py", line 936, in react
    _reactor.run()
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 1195, in run
    self.mainLoop()
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 1204, in mainLoop
    self.runUntilCurrent()
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 825, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/tcp.py", line 418, in resolveAddress
    self._setRealAddress(self.addr)
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/tcp.py", line 433, in _setRealAddress
    self.doConnect()
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/tcp.py", line 562, in doConnect
    connectResult = self.socket.connect_ex(self.realAddress)
  File "/usr/lib/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
exceptions.TypeError: an integer is required
```

This doesn't happen with other tests also run from the command line. I suspect it has something to do with the fact that the director start command is not understanding that it should also start the director (since it's needed for reporting) and therefore the socks_port being set to `None`.

hellais:
This fixes #636

### DIFF
--- a/ooni/ui/cli.py
+++ b/ooni/ui/cli.py
@@ -365,7 +365,8 @@ def runTestWithDirector(director, global_options, url=None,
                         create_input_store=True):
     deck = createDeck(global_options, url=url)
 
-    d = director.start(create_input_store=create_input_store)
+    d = director.start(create_input_store=create_input_store,
+                       start_tor=start_tor)
     @defer.inlineCallbacks
     def post_director_start(_):
         try:


darkk:
Do you want to add some smoke-test that runs `ooniprobe` binary with basic tests and checks that it had not crashed (probably, as a separate ticket) ?

I've also seen the same exception & inconclusive stracktrace in the logs while
- using tor for reporting
- AND checking tor deck
- AND having tor DirAuth being blocked (plain DROP seemed to be good enough)

I'll check if it solves that bug too after the merge.

LGTM.

hellais:
> Do you want to add some smoke-test that runs ooniprobe binary with basic tests and checks that it had not crashed (probably, as a separate ticket) ?

Yeah, this seems like a good idea, but I would add it as something separate to the unittests and maybe something that we have travis run in the form of an end-to-end test.
The reason for this is that I don't want us to have unittests become something that require many minutes to run as that will mean devs will run them less often.

Issue #635, opened 2016-10-14
Measurements should be listed from newest to older


hellais:
Currently measurements are listed from oldest to newest inside of the measurement list view (kind of by accident since we actually don't do any ordering on them).

This would involve changing how `list_measurements` works to also sort the measurements once it has listed them, see: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/measurements.py#L132.

hellais:
Done.

PR #665, opened 2016-11-02
Implement sorting of measurements in measurements tab from newest to …

hellais:
…oldest

This implements: https://github.com/TheTorProject/ooni-probe/issues/635

### DIFF
--- a/ooni/measurements.py
+++ b/ooni/measurements.py
@@ -1,4 +1,5 @@
 import json
+import operator
 
 from twisted.internet import defer
 from twisted.internet.threads import deferToThread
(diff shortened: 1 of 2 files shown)


ooni/measurements.py
```diff
@@ -139,4 +140,9 @@ def list_measurements(compute_size=False):
             measurements.append(get_measurement(measurement_id, compute_size))
         except:
             log.err("Failed to get metadata for measurement {0}".format(measurement_id))
+
+    if order is not None:
+        reverse = True if order.lower() is 'asc' else False
```
darkk: IMHO, (any) code should fail aloud on unknown input strings. I'd rather trigger the failure explicitly with sentinel objects or code like that:

```
reverse = {'asc': True, 'desc': False}[order.lower()]
```

Issue #652, opened 2016-10-25
Measuring reachability of default obfs4 bridges


Em4E:
hellais suggested that OONI could help with [our project on measurement of default Tor Browser bridges](https://www.bamsoftware.com/proxy-probe/).

Here is the set of destinations we are currently measuring: [proxy-probe.txt](https://github.com/TheTorProject/ooni-probe/files/549388/proxy-probe.txt)

This list of bridges doesn't need to be kept confidential; it consists of already deployed and soon-to-be-deployed default Tor Browser bridges; i.e., the bridges that are hardcoded in the source code. We would love to have OONI testing the reachability of these destinations from many vantage points.

For these tests, we are only doing TCP reachability (with a 30 s timeout), not a full obfs4 handshake. So if possible, that's what we would like OONI to do too. Doing a full obfs4 handshake could be interesting too, but it would be more useful for us to have simple TCP reachability working in a shorter time frame.

bassosimone:
My though on this is that, if the obfs4 handshake is reasonably low on complexity (for example if it uses either a binary or a string protocol and entails just some messages) and/or if the crypto needed to implement it is either already part of libressl or a small library (like those curve libraries), then I believe the handshake itself could be implemented as a measurement-kit test.

anadahz:
Have you seen the [Bridge reachability test]
(https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-011-bridge-reachability.md) ?

hellais:
Hey @Em4E thanks for writing up this ticket.

I think going for implementing this as a set of inputs for the tcp_connect test is the best way to got.

What I would do to implement this is the following:

* [ ] Add a tcp_connect test with the required inputs to the tor deck that we are currently shipping as part of ooniprobe (I would advise we also annotate the reports with something meaningful so that we can easily search for the ones relevant to this test).

* [ ] Update ooni-resources to provision probes with this new input and deploy an update to support this.

In going forward I think we should investigate the option of packaging obfs4proxy (together with lantern and psiphon) as a static go binary inside of ooniprobe itself so that we can do more thorough testing of it.

I am quite against the idea of having our own implementation of obfs4 (or any other CCT) technology as part of ooniprobe (or measurment-kit) as it will require a considerable amount of effort to maintain it and ensure it's exactly the same as what real users are running at a given point in time.

> Have you seen the Bridge reachability test?

[...]

PR #682, opened 2016-11-21
Add support for downloading bridge lists from ooni-resources

hellais:
This implements what is described in: #652

### DIFF
--- a/ooni/deck/store.py
+++ b/ooni/deck/store.py
@@ -91,6 +91,7 @@ class InputStore(object):
 
     @defer.inlineCallbacks
     def update_tor_bridge_lines(self, country_code):
+        from ooni.utils import onion
         in_file = self.resources.child("tor-bridges").child(
             "tor-bridges-ip-port.csv"
         )
@@ -103,7 +104,7 @@ class InputStore(object):
 
         def format_row(row):
             host, port, nickname, protocol = row
-            if protocol in ("ssh", "https", "port-444"):
+            if protocol.lower() not in onion.pt_names:
                 return "{}:{}\n".format(host, port)
             return "{} {}:{}\n".format(protocol, host, port)
 
--- a/ooni/deck/store.py
+++ b/ooni/deck/store.py
@@ -21,6 +21,28 @@ class InputNotFound(Exception):
 class DeckNotFound(Exception):
     pass
 
+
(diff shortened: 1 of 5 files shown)


ooni/deck/store.py
```diff
@@ -51,30 +73,54 @@ def update_url_lists(self, country_code):
...
+
+        data_fname = "tor-bridge-lines.txt"
+        desc_fname = "tor-bridge-lines.desc"
+        out_file = self.path.child("data").child(data_fname)
+
+        def format_row(row):
+            host, port, nickname, protocol = row
+            if protocol in ("ssh", "https", "port-444"):
+                return "{}:{}\n".format(host, port)
+            return "{} {}:{}\n".format(protocol, host, port)
```
darkk: I understand, that it's required to _parse_ `protocol` from the datafile in `setUp` (compatibility with old lists in the moment of upgrade as the list will be stale for an hour or so), but I don't understand the reason to drop `protocol` selectively in `format_row`.

ooni/deck/store.py
```diff
@@ -21,6 +21,28 @@ class InputNotFound(Exception):
 class DeckNotFound(Exception):
     pass
 
+
+def write_txt_from_csv(in_file, out_file, func, skip_header=True):
+    out_fh = out_file.open('w')
```
hellais: At this point there is actually no real valid reason for not doing what you suggest. I guess I got into the bad habbit of doing the above pattern to maintain backward compatibility with python 2.6 that didn't support multiple with statements on the same line.

I will adjust accordingly.

ooni/deck/store.py
hellais: The reason for dropping the protocol at the beginning is that we care to indicate the protocol only in the case that it's a pluggable transport type.

In other cases we should just treat it as a non pluggable transport bridge and go ahead with a regular bridge test.

darkk: I see. I'd rather reuse list of known `pluggable_transports` from `tcp_connect` to make it explicit and cross-referenced...
Also, ordinary bridges deserve tor-handshake but not tcp-handshake (I assume, you meant that too), but I see no public old-school bridge [in the TBB source](https://gitweb.torproject.org/builders/tor-browser-bundle.git/tree/Bundle-Data/PTConfigs/bridge_prefs.js).
Anyway, that's not a big deal, I got your point.

Issue #657, opened 2016-10-27
Sanitise IP address also inside of response headers


hellais:
It is the case that sometimes response headers will contain the IP address of the user. 

We should do something similar to what we do for removing the IP in the response body also for the headers. That is if we notice the IP to be in there, replace it with `[REDACTED]`.

hellais:
This is now done

Issue #605, opened 2016-09-19
web_connectivity: do not follow redirects on the local host


bassosimone:
Currently the web_connectivity test will follow redirects also when the redirect is towards private IP addresses.

This is problematic for 2 reasons:
- It can lead to false positives because obviously if a site is hosted in private ip space it will not match to the site hosted in private IP space on the control.
- It can lead to potential privacy leaks if the user is hosting something sensitive on local ip space.

darkk:
That's also true for *connected* LAN for the same reasons.

I'm unsure if it's true for all RFC1918 nets as the blockpage MAY be served using "private" server.

darkk:
That's also true for *connected* LAN for the same reasons.

I'm unsure if it's true for all RFC1918 nets as the blockpage MAY be served using "private" server.

hellais:
This is now merged.

PR #685, opened 2016-11-23
Improvements/privacy

hellais:
This fixes:
* https://github.com/TheTorProject/ooni-probe/issues/605
* https://github.com/TheTorProject/ooni-probe/issues/657

### DIFF
--- a/ooni/common/txextra.py
+++ b/ooni/common/txextra.py
@@ -198,8 +198,8 @@ class FixedRedirectAgent(BrowserLikeRedirectAgent):
             locationHeaders[0]
         )
         uri = client.URI.fromBytes(location)
-        if self.ignorePrivateRedirects and (is_private_address(uri.host) or
-                                            uri.host == "localhost"):
+        if self.ignorePrivateRedirects and is_private_address(uri.host,
+                                                              only_loopback=True):
             return response
 
         deferred = self._agent.request(method, location, headers)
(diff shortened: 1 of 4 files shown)


ooni/common/txextra.py
```diff
@@ -191,12 +197,22 @@ def _handleRedirect(self, response, method, uri, headers, redirectCount):
             response.request.absoluteURI,
             locationHeaders[0]
         )
+        uri = client.URI.fromBytes(location)
+        if self.ignorePrivateRedirects and (is_private_address(uri.host) or
+                                            uri.host == "localhost"):
+            return response
```
darkk: We've discussed this matter and decided that redirect to private address space is OK and only redirect-to-directly-connected address (including `127.0.0.0/8`) should be avoided. Also, it does not provide *security* as domain name [like](http://www.tp-link.com/us/faq-1170.html) `tplinkwifi.net` can be used in `Location` as well.

IIRC, we've seen blockpages served from a server located in a private address space, so losing ability to see those pages is sort of bad. I'll verify that.

hellais: Yes this is a very valid point.

Thanks for bringing it up.

I will allow everything except loopback to pass through in this case.

> Also, it does not provide security as domain name like tplinkwifi.net can be used in Location as well.

Yes this is a limitation of the current approach, however in order to exclude also domain based redirects, given our current implementation of the HTTP test template, we would have to either pay the price of an extra A lookup before doing any HTTP request or we would have to heavily patch twisted to hook the domain resolution.

In light of the fact that we are going to be soon transition to using measurement-kit for network testing I think it makes most sense to defer this to future work and consider this as a poormans solution.

It may be worthwhile, though, to also look for redirects to stuff that is `.local`.

hellais:
@darkk is this good now? Can we merge?

PR #690, opened 2016-11-30
Run web-connectivity using measurement-kit

bassosimone:
This is the code I am using to test measurement-kit's web-connectivity engine with ooni-probe as driver when running a deck. It helped to find, e.g., measurement-kit/measurement-kit#970.

I am not sure this could be the way in which we end up integrating ooni-probe with MK, but it's for sure one of the possible ways to do it, it's quite cheap and it works. So it helps to have it here.

To run it, you need to recompile measurement-kit as in measurement-kit/measurement-kit#969.


hellais:
Issue moved to [ooni/probe-legacy #11](https://github.com/ooni/probe-legacy/issues/11) via [**ZenHub**](https://www.zenhub.com/)

PR #692, opened 2016-11-30
Better doc

donald213:


ooni/agent/agent.py
```diff
@@ -6,9 +6,21 @@
 from ooni.agent.scheduler import SchedulerService
 
 class AgentService(service.MultiService):
+    """Contain any element to manage agent services.
+
+
+    The class above inherits to another class wich is a services container."""
```
hellais: I believe this comment is redundant.

ooni/agent/agent.py
```diff
@@ -6,9 +6,21 @@
 from ooni.agent.scheduler import SchedulerService
 
 class AgentService(service.MultiService):
+    """Contain any element to manage agent services.
```
hellais: This should read:

"The AgentService is responsible for running the ooniprobe web UI and the scheduled periodic tasks as a system daemon."

ooni/agent/agent.py
```diff
@@ -6,9 +6,21 @@
...
+    The class above inherits to another class wich is a services container."""
+
+
     def __init__(self, web_ui_port):
-        service.MultiService.__init__(self)
+        """Load configuration or made the default configuration of the service.
 
+
+        Set up the child services of the AgentService.
+        If the advanced configuration is not enabled, the page is not displayed.
```
hellais: You should specify what is meant by "advanced configuration". You probably want to say:

"If `config.advanced.disabled_webui` is set to True, then the web UI will not be started"

ooni/agent/agent.py
```diff
@@ -22,7 +34,11 @@ def __init__(self, web_ui_port):
 
 
     def startService(self):
+        """Run a service."""
```
hellais: This and the comment below are redundant.

ooni/agent/scheduler.py
```diff
@@ -35,6 +35,7 @@ class FileSystemlockAndMutex(object):
     different stacks (threads/fibers) within the same process without races.
     """
     def __init__(self, file_path):
+        """Initialize a lock for the filesystem and a lock for event driven systems."""
```
hellais: These comments are redundant. If anything I would document what `file_path` is.

ooni/ui/web/web.py
```diff
@@ -6,6 +6,11 @@
 from ooni.settings import config
 
 class WebUIService(service.MultiService):
+    """Contain any element to manage the web user interface service.
+
+
+    The class above inherits to another class wich is a services container."""
```
hellais: Maybe it's sufficient to say. "This multiservice contains the ooniprobe web user interface"

ooni/ui/web/web.py
```diff
@@ -14,16 +19,23 @@ def __init__(self, director, scheduler, port_number=8842):
         self.port_number = port_number
 
     def startService(self):
+        """Start a web user interface.
+
+        Connects a given protocol factory to the given numeric TCP/IP port and open a page."""
```
hellais: Given the fact that this is a standard interface for multiservices I think these two comments are redundant.

donald213:
Thank hellais, you have been a great help. As far as I understand I have to remove every redundant lines. To be honest I do not think there will stay a lot of my work. I save what I can. Fell free to check out my next submission. See you soon.

ooni/ui/web/web.py
donald213: What about stopService()?

hellais:
I went ahead and make some changed on top of your branch and merged it into master.

(I also squashed all your commits into 1 and edited the author, in the future you may want to setup you git so that the username is set properly. You can do so with: `git config --global user.name Donald &&  git config --global user.email donald213@users.noreply.github.com`

PR #698, opened 2016-12-06
"Vagrantfile sleeps with fishes"

bassosimone:
This removes support for Vagrant that has been quite neglected
lately in favour of Docker for which we're doing efforts to have
continuous integration (and, as such, it should work very well
because we will promptly detect breakages).

To zap Vagrant related stuff I did `grep -Rni vagrant` to search
for files where Vagrant was mentioned.

Discussed with @hellais.


hellais:
LGTM!

PR #700, opened 2016-12-07
Make refresh_deck_list less noisy

bassosimone:
Do not emit messages regarding what's going on under the hood
that may end up confusing users such as /me.

Discussed on IRC with @hellais.


hellais:
LGTM

Issue #498, opened 2016-05-10
-g option should explain that IP is not redacted from the reports


anadahz:
This info needs to be added:
- Update document to mention that doesn't apply when the -g command line option is activated.
- Update OONI related documentation
- Add info to the -g command line option when ooni-probe help is invoked

Relevant [discussion](https://github.com/TheTorProject/ooni-spec/pull/57#discussion_r62433263).

hellais:
This has been implemented.

PR #709, opened 2017-01-10
-g option should explain that IP is not redacted from the reports #498

ioef:
Based on the Report #498,  3 files have been modified.

### DIFF
--- a/data/ooniprobe.1
+++ b/data/ooniprobe.1
@@ -52,6 +52,7 @@ Disable writing to disk
 .TP
 .B \-g\fP,\fB  \-\-no\-geoip
 Disable geoip lookup on start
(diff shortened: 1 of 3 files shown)


hellais:
This is great!
Thanks a lot for it.

Do you think you could squash all these commits into one (see: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits)?

hellais:
I am going to cherry pick 47fdf34 and merge that. Thanks for this!

hellais:
Landed in master here: https://github.com/TheTorProject/ooni-probe/commit/fdd215907d0c98460daaf88f8ecf5050818624ea

Closing.

Issue #707, opened 2017-01-02
Measurements should include details of which deck generated them


hellais:
For the revamped web UI we need to know which measurement is related to which deck.

This way we are able to support filtering by deck in the measurement list view.

hellais:
Done.

Issue #706, opened 2017-01-02
We should keep track of the running or not running status of a deck


hellais:
Possibly related to #613 it should be possible to see from the UI if a deck is currently running (either because it was scheduled or because it was run manually) and if so inform the user of this.

hellais:
Done.

Issue #705, opened 2017-01-02
The /api/deck endpoint should return the list of tests in a deck


hellais:

hellais:
Done.

Issue #704, opened 2017-01-02
Adjust the /api/deck endpoint


hellais:
Currently the `/api/deck` endpoint returns something like the following:

```json
{
    'available': {
      'tor': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Tor test deck',
        'description': 'This deck runs test related to testing the reachability of the Tor network'
      },
      'web': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Web test deck',
        'description': 'This deck runs HTTP Header Field Manipulation and the Web Connectivity test'
      },
      'im': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Instant Messaging deck',
        'description': 'This test deck will check to see if instant messaging applications are working'
      },
      'http-invalid': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'HTTP Invalid Request Line',
        'description': 'This deck runs HTTP Invalid Request Line test'
      }
    },
    'enabled': {
      'tor': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Tor test deck',
        'description': 'This deck runs test related to testing the reachability of the Tor network'
      },
      'web': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Web test deck',
        'description': 'This deck runs HTTP Header Field Manipulation and the Web Connectivity test'
      },
      'im': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'Instant Messaging deck',
        'description': 'This test deck will check to see if instant messaging applications are working'
      },
      'http-invalid': {
        'schedule': '@daily',
        'enabled': true,
        'name': 'HTTP Invalid Request Line',
        'description': 'This deck runs HTTP Invalid Request Line test'
      }
    }
}
```

Instead it should return something like:

```json

[
  {
    'id': 'tor',
    'icon': 'fa-wrench',
    'schedule': '@daily',
    'enabled': true,
    'running': false,
    'name': 'Tor test deck',
    'nettests': [
      'vanilla_tor'
    ],
    'description': 'This deck runs test related to testing the reachability of the Tor network'
  },
  {
    'id': 'web',
    'icon': 'fa-unlink',
    'schedule': '@daily',
    'enabled': true,
    'running': false,
    'name': 'Web test deck',
    'nettests': [
      'http_header_field_manipulation',
      'web_connectivity'
    ],
[...]

PR #713, opened 2017-01-13
Feature/deck status

hellais:
This branch adds support for the following features:

* We should keep track of the running or not running status of a deck #706
* Adjust the /api/deck endpoint #704
* The /api/deck endpoint should return the list of tests in a deck #705
* Measurements should include details of which deck generated them #707


bassosimone:
Looks good to me. I am going to rebase locally and make sure nothing is lost / added during the process by using `git diff c849887`. If that works, I am going to force push and merge.

Issue #479, opened 2016-04-20
support unix-socket tor control ports


willscott:
The tor control port is currently hard-coded as a tcp socket
https://github.com/TheTorProject/ooni-probe/blob/master/ooni/settings.py#L212

in a debian system tor for instance, it would be preferable to connect to the unix control socket. this cannot currently be specified.

meejah:
In the `TorConfig` instance you pass in, you can specify your own `ControlPort` config and that will be respected by `txtorcon`. So, e.g.:

```
config = TorConfig()
config.ControlPort = 'unix:/var/log/tor/control`
# ... etc
```

hellais:
This is now implemented thanks to @ioef!

PR #715, opened 2017-01-15
support unix-socket tor control ports #479

ioef:
According to the Issue #479 a small number of changes have been performed in the files:
- data/ooniprobe.conf.sample
- ooni/settings.py

### DIFF
--- a/ooni/settings.py
+++ b/ooni/settings.py
@@ -538,6 +538,6 @@ class OConfig(object):
                 else:
 			incoherent.append('tor:control_port')
 
-       self.log_incoherences(incoherent)
+        self.log_incoherences(incoherent)
 
 config = OConfig()
--- a/ooni/settings.py
+++ b/ooni/settings.py
@@ -519,24 +519,24 @@ class OConfig(object):
                 except Exception:
                     incoherent.append('tor:socks_port')
 
-	    if self.tor.control_port is not None:
-       	        if isinstance(self.tor.control_port, int):
-	            control_port_ep = TCP4ClientEndpoint(reactor,
+            if self.tor.control_port is not None:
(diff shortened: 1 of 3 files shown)


ooni/settings.py
```diff
@@ -519,14 +519,19 @@ def check_tor(self):
                 except Exception:
                     incoherent.append('tor:socks_port')
 
-            if self.tor.control_port is not None:
-                control_port_ep = TCP4ClientEndpoint(reactor,
+	if self.tor.control_port is not None:
```
hellais: I think you have lost 1 layer of indentation of the if statement here.

ooni/settings.py
```diff
@@ -519,14 +519,19 @@ def check_tor(self):
...
+       	    if isinstance(self.tor.control_port,int):
```
hellais: Minor: add space before int.

ooni/settings.py
```diff
@@ -519,14 +519,19 @@ def check_tor(self):
...
-                    yield connectProtocol(control_port_ep, ConnectAndCloseProtocol())
-                except Exception:
-                    incoherent.append('tor:control_port')
+		try:
+		    yield connectProtocol(control_port_ep, ConnectAndCloseProtocol())
+		except Exception:
+		    incoherent.append('tor:control_port')		
+            else:
+		if self.tor.control_port.lstrip.startswith("unix:"):
+		    control_port_ep = self.tor.control_port.lstrip("unix:")
```
hellais: This variable should probably be called something different than `control_port_ep` (`unix_socket_path`?) and you should also check that the file path exists.
If not you should do `incoherent.append('tor:control_port')`.

hellais:
LGTM

Issue #613, opened 2016-10-01
Progress bar or expected time remaining


willscott:
Especially when running a deck interactively, it would be very helpful to be able to see progress with finer granularity than that a test is running. Especially for web-connectivity, where the test is run sequentially per-url, this can be as simple as providing a line of output as each new input is processed to indicate e.g. that this is test 100/500.

hellais:
I think this is a very good suggestion, thought it's not as simple to implement as you would think.

Currently we don't go through the whole list of URLs so we don't actually know when we start a test what is the total input count.

I think doing this is probably not too hard and I would probably do as part of this also some randomisation of the order in which URLs are tested (as per: https://github.com/TheTorProject/ooni-probe/issues/480)

bassosimone:
fyi: measurement-kit emits the progress during the test. This is currently done through the logging subsystem which receives a log message formatted as JSON and can be [processed similarly to what we do in netprobe-ios](https://github.com/measurement-kit/netprobe-ios/blob/master/NetProbe/Model/NetworkMeasurement.mm#L156) to inform the user of progress.

hellais:
This is done.

Issue #586, opened 2016-08-13
Add support for showing progress of a test run


hellais:
Currently it's not possible for a user of ooniprobe to know what is the progress of a given measurement, that is how many URLs have been tested/total number of URLs to be tested.

This is particularly annoying given the fact that when running it against fairly large URL lists the testing can take up to an hour.

re: https://lists.torproject.org/pipermail/ooni-talk/2016-August/000017.html

hellais:
This is done.

PR #719, opened 2017-01-19
Feature/test progress

hellais:
This implements: https://github.com/TheTorProject/ooni-probe/issues/613 https://github.com/TheTorProject/ooni-probe/issues/586

### DIFF
--- a/ooni/nettest.py
+++ b/ooni/nettest.py
@@ -1,7 +1,8 @@
 import os
 import re
-import time
 import sys
+import copy
+import time
 
 from twisted.internet import defer
 from twisted.python.filepath import FilePath
@@ -543,6 +544,8 @@ class NetTest(object):
 
     @property
     def completionPercentage(self):
+        if self._totalInputs == 0:
+            return 0.0
         return float(self._completedInputs) / float(self._totalInputs)
 
     @property
@@ -600,14 +603,12 @@ class NetTest(object):
     def initialize(self):
         for test_class, _ in self.testCases:
             # Initialize Input Processor
-            inputs = yield defer.maybeDeferred(
-                test_class().getInputProcessor
-            )
-            for _ in inputs:
-                self._totalInputs += 1
+            test_instance = test_class()
             test_class.inputs = yield defer.maybeDeferred(
-                test_class().getInputProcessor
+                test_instance.getInputProcessor
             )
+            if test_instance._totalInputs != -1:
+                self._totalInputs += test_instance._totalInputs
 
             # Run the setupClass method
             yield defer.maybeDeferred(
@@ -760,6 +761,8 @@ class NetTestCase(object):
 
     localOptions = {}
 
+    _totalInputs = -1
+
     @classmethod
     def setUpClass(cls):
         """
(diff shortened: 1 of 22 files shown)


bassosimone:
```
$ cat .git/refs/heads/feature/test-progress 
d2a1e98e2b178b58f46cde81b116eea5aa7f8e24
$ git diff 8c9f4ec1f49653c8b8197d67e0e4c7850b49430f
$ 
```

ooni/nettest.py
```diff
@@ -541,6 +561,13 @@ def doneReport(self, report_results):
         The state for the NetTest is informed of the fact that another task has
         reached the done state.
         """
+        self._completedInputs += 1
+        log.msg("Status")
+        log.msg("------")
+        log.msg("%d completed %d remaining" % (self._completedInputs,
+                                               self._totalInputs))
+        log.msg("%0.1f%% (ETA: %ds)" % (self.completionPercentage * 100,
+                                        self.completionEta))
```
bassosimone: There is something wrong in the way in which ETA is computed. From my console logs:

```
Status
------
4 completed 0 remaining
0.0% (ETA: -4s)
```

Note: the test here was http-invalid-request-line.

ooni/nettest.py
```diff
@@ -730,6 +761,8 @@ class NetTestCase(object):
 
     localOptions = {}
 
+    _totalInputs = -1
```
bassosimone: I think `None` is a better dummy value here

bassosimone: Or maybe zero? (That is, I don't get why -1 is used here.)

ooni/ui/web/server.py
```diff
@@ -472,7 +472,7 @@ def api_nettest_start(self, request, test_name):
                 400, 'Missing required option: "{}"'.format(option_name)
             )
 
-        except usage.UsageError:
+        except usage.UsageError as ue:
             raise WebUIError(
```
bassosimone: Where is this `ue` used?

hellais:
> the progress is not updated in the #/measurements view

By this you mean that you don't see the progress counter increasing if you just stay on the page and don't refresh? That is a known limitation and is a consequence of the fact that we don't really have a way of broadcasting updates for specific elements in the REST API to the GUI.

The ideal way to do this, I think, would be using something like websockets. An alternative would be to use an approach similar to how we send notifications from the backend to the client (long polling).

I will look again if there is an easy way to support this in the client and the backend, but otherwise I would say we defer this to future versions.

Note: the refresh button, on the right of the logs button, was added specifically for this reason and an easy way to overcome this limitation.

 > I'd add one more empty line before printing the Status text on the standard output

OK

ooni/nettest.py
hellais: Yes you are right. It was meant to be some sort of flag, but I realise now that I actually don't check it. I think at this point I should probably use None and check for None in the above.

hellais:
> the progress is not updated in the #/measurements view

I am going to consider integrating this directly into the #723 pull request. This branch here is now ready for merge.

bassosimone:
> I will look again if there is an easy way to support this in the client and the backend, but otherwise I would say we defer this to future versions.

I understand. I would probably be wise to postpone this specific issue. (Maybe it would be also worth to have a ticket, tho, such that people is not going to open a ticket for us?)

bassosimone:
> This branch here is now ready for merge.

I can't guarantee to be able to do further testing by today. I have read aa9d2a4, however, it it makes sense to me. If waiting for me to do another pass is slowing you down, please go on with merge!

hellais:
> I understand. It would probably be wise to postpone this specific issue. (Maybe it would be also worth to have a ticket, tho, such that people is not going to open a ticket for us?)

Filed a ticket for this inside of #728 

> I can't guarantee to be able to do further testing by today. I have read aa9d2a4, however, and it makes sense to me. If waiting for me to do another pass is slowing you down, please merge!

Ok I am going to proceed with merging this so I can work on tagging a release candidate.

PR #721, opened 2017-01-20
Update settings.py for the correct usage of  lstrip builtin method 

ioef:
After reviewing the code one more time i found out that the lstrip method is not used correctly.


hellais:
LGTM

PR #723, opened 2017-01-24
Release/2.2.0 rc.0

hellais:
This branch is an integration branch for the new web UI and ooniprobe 2.2.0.

It includes also some minor bugfixes found while testing the new web UI with ooniprobe.


anadahz:
`Web browser: Firefox ESR 45.7.0`

The 2nd step of the first time setup is out of the visible screen view.
ooniprobe-2 2 0rc0

Not sure if this issue should be better filled in: https://github.com/TheTorProject/ooni-wui/issues

data/decks/web.yaml
```diff
@@ -1,6 +1,12 @@
 ---
-name: Web test deck
-description: This deck runs HTTP Header Field Manipulation and the Web Connectivity test
+name: Test URLs
+description: "This deck examines whether (and how) websites are blocked and
+tries to detect censorship and/or surveillance systems in your network. The Web
+Connectivity test examines whether access to websites is blocked through DNS
+tampering, TCP/IP blocking, or by a transparent HTTP proxy. The
+HTTP Header Field Manipulation test tries to detect censorship and/or
```
anadahz: As the category name changed to 'Test URLs'  perhaps makes more sense to add the HTTP Header Field Manipulation test to 'Test for Middle Boxes' but in a different deck than http-invalid.yaml --so that we can still disable it separately.

hellais:
> The 2nd step of the first time setup is out of the visible screen view.

Can you provide a screenshot with the bug? Do you perhaps have javascript disabled?

hellais:
I tested on FF 51.0.1 (64-bit) and it works for me :|

data/decks/web.yaml
hellais: Actually there is no such thing as "category", what has changed is only the deck name (that is to say there is no way to put multiple decks inside of a certain "category").

I think maybe the best thing to do is to rename 'Test URLs' to 'Web tests'.

anadahz:
> Can you provide a screenshot with the bug?

screenshot

>  Do you perhaps have javascript disabled?

Javascript is activated.
This is a clean profile with no extensions loaded.
Distribution (debian) package version: 45.7.0esr-1~deb8u1

hellais:
@anadahz that is very strange. Do you see the animation happening (the sliding of the page to the right)?

anadahz:
@hellais This bug affects  (Windows Vista) Firefox versions 45-47 (perhaps other versions as well), you can reproduce it [here](https://www.browserling.com/browse/win/vista/firefox/45/http%3A%2F%2F37.218.242.32%3A3331).

hellais:
@anadahz ok I think I have found a way to fix this. See if http://37.218.242.109:3000/ works for you.

I have tested that it works on Ubuntu.

anadahz:
@hellais Thanks, this works for me.
BTW, is Safari browser supported?

hellais:
> BTW, is Safari browser supported?

Yes it should work without a problem in Safari. The only browser I can imagine there being issues with is IE <= 10, but also that it will probably just break the layout in some of the places where we use flexboxes and there was no cleaner way of doing it.

hellais:
I am going to proceed with merging and tagging this.

hellais:
I rebased and merged this in master.

PR #729, opened 2017-01-31
Fix bug in handling of request headers

hellais:
This bug is the reason why when we hit a redirect measurements are inconsistent.

Basically we were not passing along the headers when handling a redirect.

Weirdly enough the bug is due to us not properly implementing the `copy` method inside of our custom `TrueHeaders` class.

### DIFF
--- a/ooni/common/txextra.py
+++ b/ooni/common/txextra.py
@@ -37,6 +37,12 @@ class TrueHeaders(Headers):
         self._rawHeaders[name.lower()]['name'] = name
         self._rawHeaders[name.lower()]['values'] = values
 
+    def copy(self):
+        rawHeaders = {}
+        for k, v in self.getAllRawHeaders():
+            rawHeaders[k] = v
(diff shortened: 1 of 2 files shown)


bassosimone:
# TODO list for /me

- [ ] diff looks good?
- [ ] do I like the name? *yeah, for sure, txextra is a super beautiful name*
- [ ] have I tested it locally?
- [ ] have I written a similar test for MK?

bassosimone:
*Note*: as discussed in private with @hellais, we may want to share more information with the backend such that the backend knows whether we're dealing correctly with redirection, as it is now the case, or not, as it used to be, and behave accordingly. I think this can either be done in this or in a separate pull request. AFAIAC, up until 2751fe2 the code is approved by `/me`.

hellais:
I am going to merge this into the integration branch and eventually apply a fix to that branch directly if I decide to go down that route.

PR #730, opened 2017-01-31
Feature/http request headers

hellais:


hellais:
I am going to proceed with merging this into the integration branch

PR #739, opened 2017-03-17
Use the same logic for encoding header values and keys as measurement…

hellais:
…-kit

* Fix edge case that lead us to not do stripping of response headers
 when response body was null.

### DIFF
--- a/ooni/templates/httpt.py
+++ b/ooni/templates/httpt.py
@@ -167,19 +167,28 @@ class HTTPTest(NetTestCase):
                     getattr(response.request, 'absoluteURI', None)):
                 session['request']['url'] = response.request.absoluteURI
 
+            response_headers = {}
+            for name, value in response.headers.getAllRawHeaders():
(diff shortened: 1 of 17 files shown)


ooni/templates/httpt.py
```diff
@@ -167,19 +167,28 @@ def addToReport(self, request, response=None, response_body=None,
...
 
-            response_headers = _representHeaders(response.headers)
             # Attempt to redact the IP address of the probe from the responses
-            if (config.privacy.includeip is False and probe_ip.address is not None and
-                    (isinstance(response_body, str) or isinstance(response_body, unicode))):
-                response_body = response_body.replace(probe_ip.address, "[REDACTED]")
+            if config.privacy.includeip is False and \
+                            probe_ip.address is not None:
+                if isinstance(response_body, str) or \
+                        isinstance(response_body, unicode):
```
darkk: `isinstance(response_body, (str, unicode))` is a bit more dense.

PR #750, opened 2017-04-26
Error handling

meejah:


bassosimone:
I see that the build fails for Twisted < 15 with this error:

```
pkg_resources.UnknownExtra: Twisted 13.2.0 has no such extra feature 'tls'
```

I don't think this has anything to do with this diff, however.

I'll check whether this error was happening also before.

bassosimone:
Hmm, this [recent build of master](https://travis-ci.org/TheTorProject/ooni-probe/builds/219617541) did not fail like this. I guess I'll leave the merge to @hellais.

hellais:
I suspect the issue with:

```
pkg_resources.UnknownExtra: Twisted 13.2.0 has no such extra feature 'tls'
```
Is due to txtorcon specifying the Twisted dependency as: https://github.com/meejah/txtorcon/blob/master/requirements.txt#L4

hellais:
The build will be fixed once https://github.com/TheTorProject/ooni-probe/pull/752/commits/ef827094060e56808ebb69cd1bb69ba24c6ca458 is merged.

hellais:
This is good to be merged. Thanks for this! 👍

Issue #751, opened 2017-04-28
Test fails with `unknown_failure decoding unicode is not supported`


hellais:
### Description

I noticed that recently the whatsapp test is failing with the following error:

```
2017-04-28 01:34:53,812411+0000 [HTTP11ClientProtocol,client] [!] Failed to connect to http://149.154.171.5:443: unknown_failure decoding Unicode is not supported
```

As result this is what the result page for the whatsapp test looks like:
![screen shot 2017-04-28 at 12 13 00 0000](https://cloud.githubusercontent.com/assets/424620/25528254/5f4742c8-2c0c-11e7-9413-a0d450f6e7af.png)

### Further information

**How did you install ooniprobe**

This is running on lepidopter 1.0.0

**What operating system are you running and what version is it?**

<details><summary>Output of `ooniprobe --version`</summary>

```
ooniprobe version: 2.2.0
```
</details>

hellais:
Strangely it appears that the same version of ooniprobe (2.2.0) on macOS does not exhibit the same issue when running the whatsapp test from the command line.

anadahz:
I was able to reproduce this in lepidopter, (with ooniprobe 2.2.0):
https://explorer.ooni.torproject.org/measurement/20170427T005245Z_AS1916_Kgh3TX6DsOxHjHwtEusQylNNKogzKySGqgjbH56Aacmqnrgtmu

hellais:
So I have come to the bottom of this issue. The root cause of it is txtorcon 0.19.0 (see: https://github.com/meejah/txtorcon/issues/232).

This issue would have however been mitigated if the code in question was a bit more defensive and didn't make too many assumptions on the types of values.

Here is an explanation of what is going on exactly:

The ooni function that is failing is `representBody` (https://github.com/TheTorProject/ooni-probe/blob/master/ooni/common/http_utils.py#L5) that is making the assumption that the argument it gets passed is going to be a str, but in our case it happens to instead be a unicode.

This function is actually applied also to response headers and not only to bodies.

[...]

PR #752, opened 2017-04-28
Make the geoip lookup more robust

hellais:
related to #751


hellais:
I am going to proceed with merging this.

PR #753, opened 2017-05-03
repalcing urlparse.urlparse() by uritools

donald213:
Hey!

I replaced urlparse by uritools for a cleaner code.

If you want I replace urlparse by uritools in another module(s), fell free to ask.

If you do not think we should replace urlparse by uritools, fell free to ask.

Bye.


hellais:
It seems like you based your branch on top of something that is not master.

Also I don't think we want to be adding an extra dependency to ooniprobe (uritools) for url parsing given the fact that urlparse is fit our purposes.

donald213:
I do ot master git. Please can you give me the instructions to merge? I saw already the doc for developper.


I do notthink with you we should keep urlparse is better. Urlparse is less maintained and harder to deploy than uritools. I am not even sure that urlparse works with python3.


See you soon.

________________________________
De : Arturo Filastò <notifications@github.com>
Envoyé : mercredi 3 mai 2017 13:27
À : TheTorProject/ooni-probe
Cc : donald213; Author
Objet : Re: [TheTorProject/ooni-probe] repalcing urlparse.urlparse() by uritools (#753)


It seems like you based your branch on top of something that is not master.

Also I don't think we want to be adding an extra dependency to ooniprobe (uritools) for url parsing given the fact that urlparse is fit our purposes.

hellais:
> I do notthink with you we should keep urlparse is better. Urlparse is less maintained and harder to deploy than uritools. I am not even sure that urlparse works with python3.

urlparse is part of the standard python library, so it's as well mantained as ... python.

Currently ooniprobe does not target python3, but if we were to target python3 it's quite trivial to use urparse in a way that is compatible with both python2 and python3 (it's matter of either using `six` or doing imports in a conditional way).

hellais:
I am going to close this PR.

donald213:
Wait... Can you just be more precise about what the instruction wants to do?

________________________________
De : Arturo Filastò <notifications@github.com>
Envoyé : samedi 27 mai 2017 00:03:08
À : TheTorProject/ooni-probe
Cc : donald213; Author
Objet : Re: [TheTorProject/ooni-probe] repalcing urlparse.urlparse() by uritools (#753)


I am going to close this PR.

hellais:
> Wait... Can you just be more precise about what the instruction wants to do?

We don't want to add an extra dependency to ooniprobe.

PR #760, opened 2017-06-04
Should close #639.

### LINKED ISSUE #639: web connectivity test input without http(s):// prefix
Web connectivity test url option should either append http:// if no http(s):// is being supplied 
or inform the user that the test is not going to work for a given input (URL) that has no http(s):// prefix.

luceatnobis:
This is my first pull request to a major project, please be mindful of
that. If web_connectivity.py was issued a URL that urlparse() cannot
disseminate into individual elements, an exception will be raised.
Furthermore, if the supplied URL does not start with 'http', InvalidURL
will be raised with a message addressing this particular issue.

### DIFF
--- a/ooni/nettests/blocking/web_connectivity.py
+++ b/ooni/nettests/blocking/web_connectivity.py
@@ -25,7 +25,7 @@ from ooni.utils.net import COMMON_SERVER_HEADERS
 class InvalidControlResponse(Exception):
     pass
 
-class AbsentHostname(Exception):
+class InvalidURL(Exception):
     pass
 
 class UsageOptions(usage.Options):
@@ -164,9 +164,18 @@ class WebConnectivityTest(httpt.HTTPTest, dnst.DNSTest):
         self.report['tcp_connect'] = []
         self.report['control'] = {}
 
-        self.hostname = urlparse(self.input).netloc
-        if not self.hostname:
-            raise AbsentHostname('No hostname', self.input)
+        parseResult = urlparse(self.input)
+
+        # reject --url parameter in case http(s):// was not supplied
+        if not parseResult.scheme and not parseResult.hostname:
+            if not self.input.startswith("http"): 
+                # we want notify the user that no scheme was provided
+                raise InvalidURL('URL must contain http(s):// scheme, test will fail')
+            else:
+                # we don't know why input isn't valid
+                raise InvalidURL('Invalid URL', self.input)
+
+        self.hostname = parseResult.hostname
 
         self.control = {
             'tcp_connect': {},
--- a/ooni/nettests/blocking/web_connectivity.py
+++ b/ooni/nettests/blocking/web_connectivity.py
@@ -20,6 +20,7 @@ from ooni.common.tcp_utils import TCPConnectFactory
 from ooni.errors import failureToString
 from ooni.templates import httpt, dnst
 from ooni.utils import log
+from ooni.utils.url import validateURL
 from ooni.utils.net import COMMON_SERVER_HEADERS
 
 
@@ -27,6 +28,9 @@ class InvalidControlResponse(Exception):
     pass
 
 class InvalidURL(Exception):
+    pass
+
+class InvalidScheme(Exception):
     pass
 
 class UsageOptions(usage.Options):
@@ -175,16 +179,18 @@ class WebConnectivityTest(httpt.HTTPTest, dnst.DNSTest):
 
         parseResult = urlparse(self.input)
 
-        # reject --url parameter in case http(s):// was not supplied
-        if not parseResult.scheme and not parseResult.hostname:
-            if not self.input.startswith("http"): 
-                # we want notify the user that no scheme was provided
-                raise InvalidURL('URL must contain http(s):// scheme, test will fail')
-            else:
-                # we don't know why input isn't valid
-                raise InvalidURL('Invalid URL', self.input)
+	if parseResult.scheme == "":
+            url = "http://{}".format(self.input)
+        elif parseResult.scheme in ['http', 'https']:
+            url = self.input
+        else:
+            raise InvalidScheme("Invalid Scheme", parseResult.scheme)
 
-        self.hostname = parseResult.hostname
+        # we will leave the rest of the validation up to validateURL
+        if not validateURL(url):
+            raise InvalidURL("Invalid URL", url)
+
+        self.hostname = parseResult.netloc
 
         self.control = {
             'tcp_connect': {},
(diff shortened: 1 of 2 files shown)


ooni/nettests/blocking/web_connectivity.py
```diff
@@ -164,9 +164,18 @@ def setUp(self):
         self.report['tcp_connect'] = []
         self.report['control'] = {}
 
-        self.hostname = urlparse(self.input).netloc
-        if not self.hostname:
-            raise AbsentHostname('No hostname', self.input)
+        parseResult = urlparse(self.input)
+
+        # reject --url parameter in case http(s):// was not supplied
+        if not parseResult.scheme and not parseResult.hostname:
```
hellais: `ParseResult` doesn't have an attribute called `.hostname`.

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -164,9 +164,18 @@ def setUp(self):
...
+            if not self.input.startswith("http"):
```
hellais: Actually if you have entered in here (because `not parseResult.scheme == True`), it means that self.input will probably not start with `http` already (unless it has another SCHEMA).

hellais:
Good stuff! Thanks for working on this! 👍 

I have the following suggestions though:

The goal of #639 is to actually have the `http://` prefix be appended to domains that don't contain it. As such what we should be doing here is actually put the same logic part of the input processor (see: https://github.com/luceatnobis/ooni-probe/blob/5211f414b1aca2f19b0fe9a913b023abb0ce14c6/ooni/nettests/blocking/web_connectivity.py#L127) inside of the setup method. That way we will be fixing the case when the user specifies a hostname instead of the full URL.

luceatnobis:
It is my understanding that the user can enter a URL with either an http:// or an https:// scheme. Should it really append an http:// scheme? I'm not sure if that is a wise default, please clarify this.

Also, it would be sane to check if an invalid scheme (anything but http(s)://) was provided and in this case abort the test. Let me know what you think about this.

hellais:
> It is my understanding that the user can enter a URL with either an http:// or an https:// scheme. Should it really append an http:// scheme? I'm not sure if that is a wise default, please clarify this.

Yeah, this is actually the behaviour that we follow already when testing lists of URLs and is common behaviour for user facing UIs. Think, for example of a browser URL bar, when you type `torproject.org` it will actually fetch `http://torproject.org`.

> Also, it would be sane to check if an invalid scheme (anything but http(s)://) was provided and in this case abort the test. Let me know what you think about this.

Yes I agree we should also do some sanity checks to ensure that the URI is valid and if not throw some exception.

luceatnobis:
I'm not sure what you mean 'valid'. If urlparse manages to make sense of it, would that be enough for our purposes? Or is there a corner case I'm not seeing?

hellais:
> I'm not sure what you mean 'valid'. If urlparse manages to make sense of it, would that be enough for our purposes? Or is there a corner case I'm not seeing?

In the citizenlab test-lists I use this regexp to validate URLs:
https://github.com/citizenlab/test-lists/blob/master/scripts/lint-lists.py#L10

It's a bit more strict that using urlparse and will fail if:
* The schema is not `http` or `https`
* What follows the schema is not a valid domain name or IP

luceatnobis:
I was hesitant to put a regular expression in to validate URLs, but if it already exists I will take it. Do you mind if I take this regex, refactor it into a function and make it accessible through ooni.utils.url.validate_url or something to this effect? In this case I may change scripts/init-lists.py to use this function as well, though I may do this in a different branch/pull request. What do you think about that?

hellais:
> I was hesitant to put a regular expression in to validate URLs, but if it already exists I will take it. Do you mind if I take this regex, refactor it into a function and make it accessible through ooni.utils.url.validate_url or something to this effect?

Yeah that makes sense.

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -173,9 +177,20 @@ def setUp(self):
...
-        if not self.hostname:
-            raise AbsentHostname('No hostname', self.input)
+        parseResult = urlparse(self.input)
+
+	if parseResult.scheme == "":
+            url = "http://{}".format(self.input)
+        elif parseResult.scheme in ['http', 'https']:
+            url = self.input
+        else:
+            raise InvalidScheme("Invalid Scheme", parseResult.scheme)
```
hellais: For consistency I would actually use the same code (maybe moving it into a function called `maybe_prepend_schema`) as the one in here: https://github.com/TheTorProject/ooni-probe/pull/760/files#diff-ff2446d60c29cb02ef8f4b286402a1dfR140. Otherwise we can have unexpected results depending on whether a URL is passed via the `--url` option or from a file.

luceatnobis: Where should such a function reside? It could become part of a toolset to do things with urls in utils/url.py which I have already created. If you give your ok, I will make both instances use the suggested function.

hellais: Yeah go for it I think that is a reasonable place to put it.

ooni/nettests/blocking/web_connectivity.py
```diff
@@ -176,9 +172,13 @@ def setUp(self):
         self.report['tcp_connect'] = []
         self.report['control'] = {}
 
-        self.hostname = urlparse(self.input).netloc
-        if not self.hostname:
-            raise AbsentHostname('No hostname', self.input)
+        url = prepend_scheme_if_missing(self.input)
```
hellais: Since we are already doing this in the input_processor, to avoid doing this twice, this should instead be run only when we pass the `url` option (i.e. put this function call at line 147 `self.input = prepend_scheme_if_missing (self.localOptions['url'])`)

ooni/utils/url.py
```diff
@@ -0,0 +1,30 @@
...
+
+def validateURL(url):
+    return VALID_URL.match(url)
+
+
+def prepend_scheme_if_missing(url):
+    parseResult = urlparse(url)
+
+    if parseResult.scheme == "":
+        url = "http://{}/".format(url)
```
hellais: Actually as I think of this I think we should probably avoid adding the extra `/` at the end as we risk ending up putting double `/`s when the URL is missing the scheme, but include a path (ex. `google.com/humans.txt` -> `http://google.com/humans.txt/`)

hellais: Note: this bug was actually made by me when I wrote the function in the past

hellais:
Issue moved to [ooni/probe-legacy #13](https://github.com/ooni/probe-legacy/issues/13) via [**ZenHub**](https://www.zenhub.com/)

PR #770, opened 2017-06-12
Feature/http n https

hellais:
This add support for testing sites that are specified in https also in their http version.


bassosimone:
I tested this and it works. I will merge this as soon as travis completes successfully.

Issue #616, opened 2016-10-03
Lookup probe resolver for all tests


bassosimone:
Rather than looking up the probe resolver only for web-connectivity, do this always, because it's a piece of information that helps analysis of pretty much _any_ test.

hellais:
To provide some more context, basically we lookup the IP address of the DNS resolver of the probe in here: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettests/blocking/web_connectivity.py#L86.

This means that it will only appear inside of web_connectivity tests and not inside of every ooniprobe test. Since this is actually a fairly useful bit of information I would probably add support for doing this as part of the geoip lookup: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/geoip.py#L201 and ensure that it's always written to the report somewhere in: https://github.com/TheTorProject/ooni-probe/blob/master/ooni/nettest.py#L202

ioef:
I would be willing to undertake the handling of this issue if you don't have any objection.

bassosimone:
This is now implemented in ooni/probe-engine. All OONI incarnations will eventually use it.

PR #771, opened 2017-06-15
geoip: make ProbeIP.lookup return the dns resolver too

xrmx:
Refs #616 

Wip because tests fail but ooniprobe appear to work fine.

```
$ PYTHONPATH=$PYTHONPATH:ooni coverage run $(which trial) ooni/tests/test_geoip.py
ooni.tests.test_geoip
  TestGeoIP
    test_geoip_database_version ...                                        [OK]
    test_ip_to_location ...                                                [OK]
    test_probe_ip ...                                                   [ERROR]

===============================================================================
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
DelayedCalls: (set twisted.internet.base.DelayedCall.debug = True to debug)
<DelayedCall 0x7fa6cdc6b560 [59.9094030857s] called=0 cancelled=0 Resolver.maybeParseConfig()>
```

### DIFF
--- a/ooni/geoip.py
+++ b/ooni/geoip.py
@@ -12,6 +12,7 @@ from twisted.web import client, http_headers
 client._HTTP11ClientFactory.noisy = False
 
 from twisted.internet import reactor, defer
+from twisted.names import client as dns_client
 
 from ooni.utils import log
 from ooni import errors
@@ -198,11 +199,24 @@ class ProbeIP(object):
             self.geodata['ip'] = '127.0.0.1'
 
     @defer.inlineCallbacks
+    def dns_lookup(self, dns_discovery='whoami.akamai.net'):
+        try:
+            answers = yield dns_client.lookupAddress(dns_discovery)
+            address = answers[0][0].payload.dottedQuad()
+        except Exception as exc:
+            log.exception(exc)
+            log.err("Failed to lookup the resolver IP address")
+            address = None
+        defer.returnValue(address)
+
+    @defer.inlineCallbacks
     def lookup(self, include_ip=None, include_asn=None, include_country=None):
         if self._state == IN_PROGRESS:
             yield self._looking_up
         elif self._last_lookup < time.time() - self._expire_in:
             self.address = None
+
+        self.dns_address = yield self.dns_lookup()
 
         if self.address:
             self.resolveGeodata(include_ip, include_asn, include_country)
--- a/ooni/tests/mocks.py
+++ b/ooni/tests/mocks.py
@@ -242,3 +242,19 @@ class MockCollectorClient(CollectorClient):
 
     def isReachable(self):
         return defer.succeed(True)
+
+
+class MockDnsResponsePayload(object):
+    def dottedQuad(self):
+        return '127.0.0.1'
+
+
+class MockDnsResponse(object):
+    payload = MockDnsResponsePayload()
+
+
+def mock_dns_lookup_address():
+    response = (
+        [MockDnsResponse()],
+    )
+    return defer.succeed(response)
--- a/ooni/tests/test_geoip.py
+++ b/ooni/tests/test_geoip.py
@@ -1,8 +1,11 @@
 import os
+import mock
 import shutil
 from twisted.internet import defer
+from twisted.names import client
 
 from ooni.tests import is_internet_connected, bases
+from ooni.tests.mocks import mock_dns_lookup_address
 from ooni import geoip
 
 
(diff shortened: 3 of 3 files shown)


ooni/geoip.py
```diff
@@ -198,23 +199,36 @@ def resolveGeodata(self,
...
         if self._state == IN_PROGRESS:
             yield self._looking_up
         elif self._last_lookup < time.time() - self._expire_in:
             self.address = None
 
         if self.address:
+            dns_address = yield self.dns_lookup()
             self.resolveGeodata(include_ip, include_asn, include_country)
-            defer.returnValue(self.address)
+            defer.returnValue((self.address, dns_address))
```
hellais: You probably don't want to change the signature of the `lookup` function, but instead store the dns_address state inside of the `ProbeIP` instance (maybe `self.dns_address`?). This will avoid users of the `lookup` function risking breakage.

ooni/geoip.py
```diff
@@ -198,23 +199,36 @@ def resolveGeodata(self,
...
         else:
             self._state = IN_PROGRESS
             try:
                 yield self.askTor()
                 log.msg("Found your IP via Tor")
                 self.resolveGeodata(include_ip, include_asn, include_country)
                 self._looking_up.callback(self.address)
-                defer.returnValue(self.address)
+                dns_address = yield self.dns_lookup()
+                defer.returnValue((self.address, dns_address))
```
hellais: To avoid duplicating this in multiple occasions, you probably want to re-arrange a bit the logic so that you only call `self.dns_lookup()` in one place and change the logic by which `defer.returnValue()` is called.

hellais:
The dirty reactor is happening because the dns_client library of twisted is not so nice in cleaning up after itself (it leaves a bunch of dangling deferreds).

I don't recall what I was doing in the past when I got these to cleanup the dangling deferreds. Usually in these cases (as the docstring suggests) it's useful to set `twisted.internet.base.DelayedCall.debug = True` to investigate the root cause of the issue.

xrmx:
@hellais thanks for the hints, will take another stab at this tomorrow

ooni/geoip.py
xrmx: Sure, it looks like callers don't care about it's return value though

ooni/geoip.py
xrmx: Yeah, the plan was to make it work before making it nice :)

xrmx:
I've updated the PR with your suggestions but still am not able to sort out how to cancell the callback twisted schedule for reading the dns file config. I've tried to mock the dns call but it fails too. Any hint?

```
[ERROR]
  File "/src/ooni-probe/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1386, in _inlineCallbacks
    result = g.send(result)
  File "/home/rm/src/ooni-probe/ooni/tests/test_geoip.py", line 39, in test_probe_ipdns_address
    with mock.patch.object(client, 'lookupAddress', side_effect=mock_dns_lookup_address()):
  File "/src/ooni-probe/venv/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1527, in unwindGenerator
    "caught returnValue being used in a non-generator" % (f,))
exceptions.TypeError: inlineCallbacks requires <function mock_dns_lookup_address at 0x7f4468a5b8c0> to produce a generator; insteadcaught returnValue being used in a non-generator
```

ooni/tests/mocks.py
```diff
@@ -242,3 +242,20 @@ def isSupported(self):
...
+    def dottedQuad(self):
+        return '127.0.0.1'
+
+
+class MockDnsResponse(object):
+    payload = MockDnsResponsePayload()
+
+
+@defer.inlineCallbacks
+def mock_dns_lookup_address():
```
hellais: The issue you have in the PR is in this function.

The decorator `defer.inlineCallback` can only be used when the function is a generator (i.e. it should `yield` something).

In this case since you don't actually need to "wait" on any deferred and just want to return immediately the mocked value you should change this to just be a function returning an already called deferred, like so:

```
# remove the defer.inlineCallbacks decorator
def mock_dns_lookup_address():
    return defer.succeed(
       ([MockDnsResponse()], )
    )
```

xrmx: Ah! thanks for the hint 😊

bassosimone: Ah, cool, great to know! I was also a bit confused by this when helping @xrmx to review this.

xrmx:
Updated the PR

ooni/tests/mocks.py
xrmx: Had to update the test code a bit but fortunately made it more simple. As the original function return a generator and test expect that returning a defer.succeed didn't work so i've just made it a plain python generator that works fine.

ooni/tests/mocks.py
```diff
@@ -242,3 +242,19 @@ def isSupported(self):
...
+
+class MockDnsResponse(object):
+    payload = MockDnsResponsePayload()
+
+
+def mock_dns_lookup_address():
+    response = (
+        [MockDnsResponse()],
+    )
+    yield response
```
hellais: you must not yield here, but instead return an already fire deferred via `return defer.succeed(response)`

hellais:
> Had to update the test code a bit but fortunately made it more simple. As the original function return a generator and test expect that returning a defer.succeed didn't work so i've just made it a plain python generator that works fine.


If you make it into a python generator you will not see the error anymore because you are basically creating a generator and waiting on it's yield function, but since you don't have anything that consumes the generator (ex. the `@defer.inlineCallbacks` decorator) it will block and not do anything.

ooni/tests/mocks.py
xrmx: But the code where this is called it's yielding and defer.inlineCallbacks decorated. I've tried that but IIRC it breaks with not handling StopIteration.

hellais: That's strange, can you push the branch with that fix and the `StopIteration` being raised? There is likely something decorated with an `inlineCallbacks` too many.

xrmx: Pushed!

hellais:
Issue moved to [ooni/probe-legacy #14](https://github.com/ooni/probe-legacy/issues/14) via [**ZenHub**](https://www.zenhub.com/)

PR #772, opened 2017-06-25
Use `onion.find_tor_binary()` to obtain the path to the tor binary al…

hellais:
…ways

* Avoid setting `tor_details` globally as this could be called when the
  config file has not yet been parsed and therefore it may not respect
  `advanced-tor_binary`


hellais:
I am going to close this as it's been sitting here for a while and it's a useful fix to have.

Issue #773, opened 2017-06-26
Detect carrier-grade NAT (CGNAT) addresses as private


ivilata:
### Description

The ``ooni.common.ip_utils.is_private_address()`` function detects carrier-grade NAT ([RFC6598](https://tools.ietf.org/html/rfc6598)) addresses ``100.64.0.0/10`` as public, when they should be considered private.

### Steps to Reproduce

```
$ python
>>> import ooni.common.ip_utils as ipu
>>> ipu.is_private_address('127.1.2.3')
True
>>> ipu.is_private_address('8.8.8.8')
False
>>> ipu.is_private_address('10.1.2.3')
True
>>> ipu.is_private_address('100.64.0.1')
False
```
**Expected behavior:**

```
>>> ipu.is_private_address('100.64.0.1')
True
```

### Further information

OONI probe HEAD 3063f04f with Python 2.7.13 on Debian Sid.

bassosimone:
We're doing the right thing in OONI Probe Engine: https://github.com/ooni/probe-engine/blob/13b6db5059c01bd779901755d0eaad31d7b5432f/netx/resolver/bogon.go#L17.

PR #784, opened 2017-12-20
fix is_private_address for carrier-grade NAT addresses

nmatt0:
Fix for #773 

Unfortunately python3 added an is_global function to the ipaddr class that didn't get ported to python2.

python3's is_global function:
https://github.com/python/cpython/blob/master/Lib/ipaddress.py#L1333-L1334
https://github.com/python/cpython/blob/master/Lib/ipaddress.py#L1583-L1600

I did my best to mimic python's is_global logic in my solution.


hellais:
Issue moved to [ooni/probe-legacy #15](https://github.com/ooni/probe-legacy/issues/15) via [**ZenHub**](https://www.zenhub.com/)

PR #785, opened 2017-12-29
Feature/docker

dasrecht:
Hi there 👋 

As discussed during the chaos communication congress I used docker to deploy an ooni-probe.

I've added a bit of documentation around it, the only thing left would be adding this repository to the  docker hub on the openobservatory organisation https://hub.docker.com/u/openobservatory/

The way I currently run will not persist any data on the host so every container start will start from scratch.

/b

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [run]
 source = ooni
 [report]
-omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/* Dockerfile docker-compose.yml
+omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/* Dockerfile docker-compose.yml README.rst
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [run]
 source = ooni
 [report]
-omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/* Dockerfile docker-compose.yml README.rst
+omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [run]
 source = ooni
 [report]
-omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/*
+omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/* Dockerfile docker-compose.yml
--- a/README.rst
+++ b/README.rst
@@ -167,8 +167,8 @@ Configuring ooniprobe
 ---------------------
 
 After successfully installing ooniprobe you should be able to access the web UI
-on your host machine at <http://localhost:8842/> after running:: 
-  ooniprobe -w 
+on your host machine at <http://localhost:8842/> after running::
+  ooniprobe -w
 or starting the daemon.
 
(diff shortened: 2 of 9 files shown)


hellais:
Re-starting travis to see if now that the klein patch is merged, the CI will succeed.

.coveragerc
```diff
@@ -1,4 +1,4 @@
 [run]
 source = ooni
 [report]
-omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/*
+omit = ooni/nettests/*, ooni/api/*, ooni/kit/*, ooni/contrib/* Dockerfile docker-compose.yml README.rst
```
hellais: I think you don't need to add text files to test `.coveragerc`. Since those are not source files, it will already be ignored by `coverage.py`.

README.rst
```diff
@@ -91,6 +91,28 @@ that you have the ``universe`` repository enabled. The ``universe`` repository
 is enabled by default in a standard Ubuntu installation but may not be on some
 minimal, or not standard, installations.
 
+On Docker::
```
hellais: I think you have an extra `:` in here.

README.rst
```diff
@@ -91,6 +91,28 @@ that you have the ``universe`` repository enabled. The ``universe`` repository
...
+
+Required dependencies installed on the target system:
+  - Docker
+  - docker-compose
+
+You can pull the image from our Docker Hub
```
hellais: I think you need to put at the end of the line `::` so that it considers this a code block (as opposed to inline code)

README.rst
```diff
@@ -91,6 +91,28 @@ that you have the ``universe`` repository enabled. The ``universe`` repository
...
+
+        docker pull openobservatory/ooni-probe:latest
+        docker run -p 8842:8842 -d openobservatory/ooni-probe:latest
+
+If you prefer to run the ooni-probe in a more persistent manner on your host
+you can add `--restart always` to the execution.
```
hellais: Here it's the same as above.

README.rst
```diff
@@ -91,6 +91,28 @@ that you have the ``universe`` repository enabled. The ``universe`` repository
...
+
+        docker run -p 8842:8842 --restart always -d openobservatory/ooni-probe:latest
+
+On the other hand you can also build the image on your local machine:
+
```
hellais: Here you are missing an extra `:`

docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+version: '2'
```
hellais: I think given that it's just a single service we don't really need a `docker-compose.yml` file. We can just use plain docker or wrap the commands above in a shell script.

README.rst
```diff
@@ -91,6 +91,28 @@ that you have the ``universe`` repository enabled. The ``universe`` repository
...
+        git clone https://github.com/TheTorProject/ooni-probe.git
+        docker-compose build
+        docker-compose up
```
hellais: Can you replace these two commands with just vanilla docker (without docker-compose) and remove the docker-compose dependency in the above lines?

.coveragerc
dasrecht: makes sense. will add a commit for that

dasrecht:
@hellais thanks for the review, updated as per your comments :)

hellais:
Issue moved to [ooni/probe-legacy #16](https://github.com/ooni/probe-legacy/issues/16) via [**ZenHub**](https://www.zenhub.com/)

PR #786, opened 2017-12-30
fix(travis): use klein 16.12 with old twisted

bassosimone:
It was reported on Slack that [travis was broken](https://travis-ci.org/TheTorProject/ooni-probe/builds/322884022). It seems to be caused by a new klein release requiring Twisted >= 15.5.

Use the same hack we use for other packages to force an old klein when Twisted < 15.5.


dasrecht:
👍 looks good :) this will also unbreak a few PR's which have failing travis tests

PR #789, opened 2018-01-13
Add 13067 HTTPS URLs from Chromium HSTS list

90143:
HTTPS URL input list that includes 13067 enties from
Chromium HSTS preloaded list:
https://cs.chromium.org/codesearch/f/chromium/src/net/http/
transport_security_state_static.json?cl=0


hellais:
Hi, thanks for your interest in contributing to OONI Probe!

I think that the OONI Probe `var` directory (or probably OONI Probe itself) is the right place for you to place those files in though. That directory is mainly an "attic" of old stuff that still had some sort of semi-usefulness.

PR #798, opened 2018-10-01
Fixes broken links

frankcash:
A few of the internal github links were broken, this updates that!


hellais:
Good call. I was assuming that relative links would work, but apparently they don't.

Thanks!

